Issue #29, opened 2015-03-31
Issue with path parameter encoding when parsing


gbaird:
Have an exception coming from the parser regarding URL encoding for inline parameters:

java.io.IOException: Server returned HTTP response code: 500 for URL: http://localhost:8080/Walletron/api/api-docs/ver1/brands/{brand}/templates
at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1626)
at io.swagger.parser.util.RemoteUrl.urlToString(RemoteUrl.java:46)
at io.swagger.parser.SwaggerCompatConverter.readDeclaration(SwaggerCompatConverter.java:367)
at io.swagger.parser.SwaggerCompatConverter.read(SwaggerCompatConverter.java:104)
at io.swagger.parser.SwaggerParser.read(SwaggerParser.java:34)
at io.swagger.parser.SwaggerParser.read(SwaggerParser.java:14)
at io.github.robwin.swagger2markup.Swagger2MarkupConverter$Builder.(Swagger2MarkupConverter.java:88)
at io.github.robwin.swagger2markup.Swagger2MarkupConverter.from(Swagger2MarkupConverter.java:49)
at io.github.robwin.swagger2markup.Swagger2MarkupConverterTest.testSwagger2AsciiDocConverter(Swagger2MarkupConverterTest.java:18)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:606)
...

The server is expecting the encoded URL, in this case: http://localhost:8080/Walletron/api/api-docs/ver1/brands/%7Bbrand%7D/templates That works just fine and it appears browsers like Chrome and cUrl utilities like Postman do this automatically. I am able however to reproduce the error on a straight curl from the command line.

I'm using 1.0.0

PR #31, opened 2015-03-31
Encoded URL on parse for embedded path parameters

gbaird:
See issue #29


fehguy:
Looks like this is failing tests.  I think you need to decode it in the Resolve class

gbaird:
Going to close and try another

PR #32, opened 2015-03-31
URL path encoding

gbaird:
Following the guidance here: http://stackoverflow.com/questions/724043/http-url-address-encoding-in-java

Addressing issue #31


webron:
merged to master...

Issue #39, opened 2015-04-17
Support for parsing a String into the Swagger model.


RobWin:
Hello,

it would be great to have support for parsing a String (Swagger JSON/YAML) into the Swagger model without the need to store it in a file or retrieve it via an URL.

I would like to reuse the Json.mapper() and Yaml.mapper(), but don't want to have the internal knowledge of the Swagger parser in my code.

Kind regards,
Robert

fehguy:
Can you explain more?  You mean you want to parse JSON into a model without using the swagger parser logic?

RobWin:
No. I have a Swagger Json response as a String and want to parse it into the Swagger model without having to use Json.mapper and Yaml.mapper directly.

I would like to invoke
```
Swagger swagger = new SwaggerParser().parse(jsonAsString);
```

Kind regards,
Robert

RobWin:
Something like

```
    public static SwaggerfromString(String swagger) throws IOException {
        Validate.notEmpty(swagger, "swagger must not be null!");
        ObjectMapper mapper;
        if(swagger.trim().startsWith("{")) {
            mapper = Json.mapper();
        }else {
            mapper = Yaml.mapper();
        }
        JsonNode rootNode = mapper.readTree(swagger);

        // must have swagger node set
        JsonNode swaggerNode = rootNode.get("swagger");
        if(swaggerNode == null)
            throw new IllegalArgumentException("Swagger String is in the wrong format");

        return new mapper.convertValue(rootNode, Swagger.class);
    }
```

fehguy:
I see, yes, there can be a convenience function to allow a simple string be passed in.  What you put looks fine, would you like to send a pull request for it?

RobWin:
Okidoki

PR #40, opened 2015-04-21
PR for swagger-api/swagger-parser#39

RobWin:
SwaggerCompatConverter method is not implemented yet.


fehguy:
The compat converter support isn't easily done since swagger 1.2 specs typically live in multiple files.

RobWin:
Do you accept the PR then?

fehguy:
Suggest you throw an exception as "not implemented" for 1.2 instead of failing silently then I can merge.

RobWin:
Shall I remove the new method from SwaggerParser then, if it is not supported for Swagger 1.2?
And just leave it in Swagger20Parser?

RobWin:
New method does not invoke other extensions except Swagger20Parser now.
It only supports swagger 2.0

RobWin:
@fehguy ? :-)

fehguy:
apologies for the delay

RobWin:
Thx

PR #54, opened 2015-05-25
If the spec file provides the basePath, use it.

fhriley:
The Swagger spec includes a basePath field:

https://github.com/swagger-api/swagger-core/wiki/API-Declaration/e22da2fb334170c7676cdde6e89b966e7604de13

The basePath could be different than the location of the API spec file, which will cause the parser to fail. This commit fixes that by using the basePath if it is provided and falling back to the spec file location if it is not.


fehguy:
can you please rebase?

fehguy:
The problem with doing this is, many people want to run the parser against the URL given, even if the public URL is different.  I am going to close (it's stale too) but we can revisit this if the need arises again.

Issue #59, opened 2015-06-12
SwaggerParser.parse(String swaggerAsString) always returns null


nmcb:
As it delegates to `SwaggerParser.parse(String swaggerAsString, List<AuthorizationValue> auths)` passing a `null` value as the second parameter which leads to the conditional statement in this method to always be `false`.

```java
public Swagger parse(String swaggerAsString) {
    return parse(swaggerAsString, null);
}

public Swagger parse(String swaggerAsString, List<AuthorizationValue> auths) {
    Swagger output;
    try {
        output = new Swagger20Parser().parse(swaggerAsString);
        if (output != null && auths != null && auths.size() > 0) {
            return new SwaggerResolver().resolve(output, auths);
        }
    } catch (IOException e) {
        // continue;
    }
    return null;
}
```

olensmar:
Fixed in develop branch!

PR #64, opened 2015-06-29
#59. Fixed getting Swagger from a String.

MDoumin:
Fixed parsing the swagger object from a string, when AuthList is not necessary.

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java
@@ -47,19 +47,30 @@ public class SwaggerParser {
         return null;
     }
 
+    private static Swagger getSwagger(String swaggerAsString) {
+        try {
+            return new Swagger20Parser().parse(swaggerAsString);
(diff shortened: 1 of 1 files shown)


modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java
```diff
@@ -47,19 +47,30 @@ public Swagger read(String location, List<AuthorizationValue> auths, boolean res
         return null;
     }
 
+    private static Swagger getSwagger(String swaggerAsString) {
```
olensmar: It seems to me there is one parse method too many - why not just put this code inside the parse method below?

tomtit:
I don't understand this fix at all...
I would expect to have the same logic for `auth` handling here
https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java#L28
and here
https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java#L58

PR #65, opened 2015-06-30
Fixed #59: SwaggerParser.parse(String swaggerAsString) always returns null

### LINKED ISSUE #59: SwaggerParser.parse(String swaggerAsString) always returns null

iushankin:


olensmar:
thank you!

PR #76, opened 2015-08-14
Support loading files from the classpath if they don't exist on the "filesystem"

Helmsdown:
With this update Swagger20Parser (and SwaggerResolver) will check to see if a relative file path exists on the file system and if not will try to load it from the classpath. This allows you to bring in spec files you want to reference in a jar (if you are in a Java project).

Additionally, I noticed that some Java TestNG tests were being omitted so I copied the surefire plugins block from swagger-core and that seemed to fix it.

Finally, I rewrote the RemoteUrl unit tests in Java/TestNG. They now use [Wiremock](http://wiremock.org/) to host files on a transient localhost webserver. This should remove the dependency on petstore.swagger.io being up (assuming no other tests depend on it) and @fehguy should have one less repetitive task to perform.

As always feedback on my code is appreciated.


fehguy:
looks good

PR #77, opened 2015-08-18
When loading from the classpath look in more than one place

Helmsdown:
I gave my new classpath loading a whirl when deployed in Tomcat and the ClasspathHelper did not find the swagger.json in WEB-INF/classes inside my war. 

After some googling I ran across this [this](http://stackoverflow.com/questions/7615040/difference-between-classloader-getsystemresourceasstream-and-getclass-getresou) stack overflow article. Where they describe that in a web app world the classloader situation is more complicated.

So my solution is to change the code try looking for the resource in two places.


fehguy:
yes, we've done that inside swagger-codegen as well.

PR #78, opened 2015-08-20
Nested References

rstanton:
Hi

We've just started using Swagger to define a set of commerce API (contract first) and were having a lot of trouble with nested in line references.

This update patch seems to work for us. Let me know if you need anything else as I couldn't find much info on how to contribute.

thanks

Ross


rstanton:
Realise now that 'nested inline' references makes no sense! I mean nested external references

fehguy:
Can you share a sample specification and some tests?  I think I know what you mean and would love to support this.

rstanton:
Done. Tests committed. Sample spec in here https://github.com/rstanton/SwaggerExternalRefsSamples, can put into test/resources if required

PR #89, opened 2015-09-11
Allow path (and potentially responses) for internal references, too.

ePaul:
This works together with [#1405 of swagger-core](https://github.com/swagger-api/swagger-core/pull/1405). Don't merge this one before that one, it doesn't work.

I added an internal class with 4 anonymous subclasses, one for each type of reference.
The right one is looked up by expected result type.

Maybe this is overengineered, and we should use a simple if-else?

This does not yet work for responses, because the Swagger object has no getResponses() method (or anything similar) (mentioned in [swagger-core issue #1388](https://github.com/swagger-api/swagger-core/issues/1388).

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
@@ -1,10 +1,15 @@
 package io.swagger.parser;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


ePaul:
This actually doesn't work yet with responses (throwing an UnsupportedOperationException). It might work with paths, I didn't try this yet.

fehguy:
This has been added in https://github.com/swagger-api/swagger-parser/commit/f4bb680744a79402fe1f737a1bad569cb045a5e4

PR #91, opened 2015-09-14
URL encode path segments when getting remote swagger specs

Helmsdown:
Today I came across an interesting use case that broke SwaggerParser's RemoteUrl class.

Given this example swagger 1.2 spec:
```
{
"apiVersion": "1.0.0",
"swaggerVersion": "1.2",
"apis": [
  {
    "path": "/v1/things"
  }
  {
    "path": "/v1/things/{thingId}/widgets"
  }]
}
```

RemoteUrl would get a 400 when it tried to get access: http://hostname.com/v1/things/{thingId}/widgets because the "{" and "}" characters were not url encoded. 

If you take the URL and pop it into chrome, Chrome automatically encodes the "{" and "}" characters and goes along its merry way.

I added some logic to RemoteUrl to split apart the url and only encode the path segments (e.g. ignore the  "http://hostname:port" part of the URL.

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/util/RemoteUrl.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/util/RemoteUrl.java
@@ -164,13 +164,4 @@ public class RemoteUrl {
     static {
         disableSslVerification();
     }
-
-    public static void main(String[] args) {
-        try {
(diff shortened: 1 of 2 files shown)


fehguy:
The bummer is, and this is part of the changes with the 2.0 spec, that the path `/v1/things/{thingId}/widgets` is supposed to be the "documentation path", so having `{thingId}` is actually meaningless.

Now, that doesn't mean that people don't use it differently, but that's the intent.  I don't see your changes as being harmful, and they may be helpful to others, but I wanted to make that clear.  I don't know what the actual documentation would be for that path.

Issue #93, opened 2015-09-16
HTTP 400 when following URLs with path parameters in swagger doc


ludyte:
Using swagger-codegen to read a swagger spec that aggregates other specs with path parameters in the URL, an HTTP 400 is raised because the '{' '}' that delimit the path parameter are not properly encoded by io.swagger.parser.util.RemoteUrl.   

This occurs when the Swagger parser determines that the relative path to a resource is described by a string that looks like a URL and it tries to read the contents of the URL into the parser.

```
java.io.IOException: Server returned HTTP response code: 400 for URL: http://127.0.0.1:8075/api/api-docs/v1/service/{parameter}/otherapi
    at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1838)
    at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1439)
    at io.swagger.parser.util.RemoteUrl.urlToString(RemoteUrl.java:100)
    at io.swagger.parser.SwaggerCompatConverter.readDeclaration(SwaggerCompatConverter.java:451)
    at io.swagger.parser.SwaggerCompatConverter.read(SwaggerCompatConverter.java:133)
    at io.swagger.parser.SwaggerParser.read(SwaggerParser.java:36)
    at io.swagger.codegen.Codegen.main(Codegen.java:85)
```

webron:
Are you referring to a 1.2 definition?

ludyte:
yes.   The stack trace shows the exception being raised from the compat parser.  I have a patch that's working for me, though its a bit of a hack.

ludyte:
Atleast it was until I rebased.   hack hack hack..

webron:
Yup, the problem is that it's not really a valid 1.2 definition. There's already a pending PR for a workaround, we'll see which one gets merged.

ChrLipp:
May I ask what's the state of this? I encounter the same error as ludyte and I am working currently with his PR code which works fine. I was surprised that parsing URLs with path parameters does not work with the generator.

fehguy:
I don't quite see why this is an issue.  As @webron said, it's not even a valid definition in the 1.2 specification.

Think of it like this.  In 1.2 (and prior), there was a Resource Listing, which held pointers to Api Declarations.  There is no reason why you would have a path parameter in the location of an api declaration, so there is no reason why the parser code would try to substitute a value for it.  What value would it even substitute with?

[...]

PR #94, opened 2015-09-22
HTTP 400 when following URLs with path parameters in swagger doc

ludyte:
https://github.com/swagger-api/swagger-parser/issues/93

Filters path elements for path parameters and URL encodes the braces.
Each token has to be decoded to avoid re-encoding an already encoded url.


fehguy:
This has been added in https://github.com/swagger-api/swagger-parser/commit/51e2861cb71afe69e0e030575e2b85aaef7d06ba

PR #97, opened 2015-09-23
RefUtils class constants had moved to RefType.

chackney:
Was building against the latest snapshot version 1.5.4-snapshot of core, and this class didn't compile.

HTH


webron:
Have you seen #96? :)

chackney:
Nope :), missed that. always slightly too late, out by 1 :)

webron:
Thanks for looking out though. Not sure I'll manage to merge and publish a snapshot today, but will do my best to handle it tomorrow.

PR #110, opened 2015-10-13
Fix "required" attribute during deserialization

wing328:
To fix https://github.com/swagger-api/swagger-codegen/issues/1382


fehguy:
This has been pushed to snapshots

wing328:
@fehguy thanks. I also just added test cases and will submit another PR soon

PR #117, opened 2015-10-15
Use BasicAuthDefinition for basic securityScheme

xiwenc:
The parser was incorrectly mapping a BasicAuthentication SecurityScheme to ApiAuthentication. For instance with this fix proper java client is generated with support for Basic Auth:

`client/src/main/java/com/example/service/ApiClient.java`:
```java
  public ApiClient() {
    // Use ISO 8601 format for date and datetime.
    // See https://en.wikipedia.org/wiki/ISO_8601
    this.dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");

    // Use UTC as the default time zone.
    this.dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));

    // Set default User-Agent.
    setUserAgent("Java-Swagger");

    // Setup authentications (key: authentication name, value: authentication).
    authentications = new HashMap<String, Authentication>();
    authentications.put("hallo", new HttpBasicAuth());
    // Prevent the authentications from being modified.
    authentications = Collections.unmodifiableMap(authentications);
  }

```

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
@@ -52,6 +52,7 @@ import io.swagger.models.Swagger;
 import io.swagger.models.Tag;
 import io.swagger.models.Xml;
 import io.swagger.models.auth.ApiKeyAuthDefinition;
+import io.swagger.models.auth.BasicAuthDefinition;
 import io.swagger.models.auth.In;
 import io.swagger.models.auth.OAuth2Definition;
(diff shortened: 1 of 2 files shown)


djvdorp:
Nice find! I also hit this exact same issue earlier this week.

wing328:
@xiwenc thanks for the PR. would be nice if you can add a test case to cover the change.

(FYI. https://github.com/swagger-api/swagger-parser/pull/115 has a couple of new test cases)

xiwenc:
@wing328 Good idea! I've added coverage for that piece now.

wing328:
@xiwenc :+1:

PR #121, opened 2015-10-25
handle `BasicAuthorization` in `SwaggerCompatConverter`

### LINKED ISSUE #120: basicAuth is not handled in `SwaggerCompatConverter`
As I can see, there is a similar `securityDefinition` whose `type` is `basic` in latest swagger(swagger 2.0), but at [here](https://github.com/swagger-api/swagger-parser/blob/18bd0c6a8c990260d0f35c327796199df9b56e7a/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java#L590), both `OAuth2Authorization` and `ApiKeyAuthorization` are handled but nor `BasicAuthorization`.

Is there any reason for which `BasicAuthorization` is ignored?

If none, I will try to fix this problem.

huiyiqun:
try to fix #120 .

Please tell me if there are some existing tests, and I will update tests later.

### DIFF
--- a/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
+++ b/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
@@ -654,7 +654,6 @@ public class SwaggerCompatConverter implements SwaggerParserExtension {
 
                     swagger.securityDefinition(authNickname, def);
                 } else if (auth instanceof BasicAuthorization) {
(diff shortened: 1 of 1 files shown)


huiyiqun:
Why nobody care this?

huiyiqun:
Is there anybody could help me?

fehguy:
Many projects going now so can't always jump on these

huiyiqun:
Thanks

fehguy:
this has been pushed to 1.0.14-SNAPSHOT

PR #128, opened 2015-10-29
Fix discriminator.

arnested:
This looks like an error - probably copy'n'paste gone wrong.


arnested:
Thank you for the quick merge!

Is there any chance a (snapshot) release could be pushed to [Sonatype](https://oss.sonatype.org/#nexus-search;quick~swagger-parser)?

webron:
Pushed.

arnested:
Thank you!

Issue #135, opened 2015-11-09
How to make relative refs work?


samskiter:
If I have the following folder structure:

```
./schemas
    - contents.schema.json
    - /room
        - cupboard.schema.json
        - shelf.schema.json
```
Where contents has a ref to `./room/cupboard.schema.json`
How do I add a ref inside cupboard to shelf?

`./shelf.schema.json` does work if I have an endpoint that uses cupboard directly. But doesn't if I have another endpoint that uses contents.schema.json?

samskiter:
I dug through the code a bit to try and work out what's going on. The `ResolverCache` is initialised with a parent directory that it always uses. I've got to a point where I have the directory it should be used passed in through `loadRef` and resolved all the compile errors associated (a change there propagates quite a way!). It should also fix a related issue which is that relative URL refs don't seem to be supported.
No tests currently pass though and I think a much better way to do this would be to change the Model and Swagger classed to have a `location` property that allows any refs contained within to be resolved better. If you'd like me to share where I'm at with this, I can submit a PR for your amusement.

fehguy:
this has been added as of 1.0.18

PR #136, opened 2015-11-09
Make external relative refs work

samskiter:
This is a WIP that you are welcome to make additions to. I'm not sure how far I'll be able to get to getting all the tests working but I think the core idea is there. See https://github.com/swagger-api/swagger-parser/issues/135


samskiter:
Hello, is it worth continuing this PR? Is it something you want?

fehguy:
I don't know what the goal is here other than making relative refs work.  It also fails the tests, so I don't really know what state the PR is in?

samskiter:
The aim is to make relative references work yes.

The state is that the non-test code has no compile errors and should be approximately correct (obviously tests should reveal errors). I have yet to get round to finish updating all the test cases - they will take me some time though so if you could look over what I have so far and say if you'd like me to continue or approach the problem differently...

fehguy:
This has been addressed in https://github.com/swagger-api/swagger-parser/commit/f4bb680744a79402fe1f737a1bad569cb045a5e4

PR #165, opened 2016-01-08
Don't require properties attribute on schema definitions

magnushiie:
Currently the following snippet is parsed as invalid with message "attribute definitions.CssColor.properties is missing".
```yaml
definitions:
  CssColor:
    pattern: ^#[0-9A-F]{6}$
    type: string
```

The `properties` attribute is not relevant for schemas that are not of `type: object`, and even for object schemas, the [JSON schema spec](http://json-schema.org/latest/json-schema-validation.html#anchor64) states that it can be missing and the default value is empty object.

This PR removes the requirement for the `properties` attribute.


macrosak:
:+1: Any plans to merge this?

magnushiie:
I would also greatly appreciate merging this or some feedback.

fehguy:
Sorry for the delay.  Yes, this is just fine.

Issue #169, opened 2016-01-23
SSL-error when reading swagger definition 


olensmar:
The following code:

```
        SwaggerParser parser = new SwaggerParser();
        final Swagger swagger = parser.read("https://api.rocrooster.net/api-docs.json");
```

Will fail with an SSL-related error - I've created a patch in branch https://github.com/swagger-api/swagger-parser/tree/httpclient-for-ssl which uses a lax:ed HttpClient for reading the definition - after which the above code works fine (it's added as a simple unit test in the branch)

Not sure if you wanted to introduce the httpclient dependency in the core parser - so I haven't done a PR until you've had a look

frantuma:
@fehguy @olensmar An alternative approach possibly limited to the specific or similar issue (call to https://api.rocrooster.net/api-docs.json with server checking user agent) without adding deps, would be adding User-Agent header to RemoteUrl connection, as in #171

olensmar:
thanks - the original issue that spurred this is at http://community.smartbear.com/t5/SoapUI-NG/How-to-create-SoapUI-NG-project-from-Swagger-over-HTTPS/m-p/112463#U112463 - but since we don't have access to their swagger-definition it's hard to know if neither efforts would help... I would prefer to "shoot broadly" and go with the http-client solution - but it's your call...

olensmar:
ok - their Swagger definition actually seems to work fine with older versions of the parser now - so I'm changing my mind and would be fine with just adding a User-Agent header to the RemoteUrl connection as suggested by @frantuma

fehguy:
Ok, parser 1.0.17 was released yesterday, can you update to that?

olensmar:
will do - thanks

olensmar:
I still get 

```
reading from https://api.rocrooster.net/api-docs.jsonjavax.net.ssl.SSLException: Received fatal alert: internal_error
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:208)
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:154)
	at sun.security.ssl.SSLSocketImpl.recvAlert(SSLSocketImpl.java:2011)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1113)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1363)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1391)
[...]

PR #171, opened 2016-01-26
Refs #169 - User agent in URLConnection used by Swagger20Parser.

frantuma:
@fehguy 
Alternative approach for #169, possibly limited to the specific or similar issue (call to https://api.rocrooster.net/api-docs.json with server checking user agent) without adding deps; 

adds User-Agent header to RemoteUrl connection

tested with https://api.rocrooster.net/api-docs.json in https://github.com/frantuma/swagger-parser/commit/a4a1e8bb3dea8493a1fcaacaf2d97b590f13a08b, test removed because hitting external URL


fehguy:
OK great--so the issue with this service was simply the lack of user agent?  That looks good to me, I might suggest setting it to Apache-HttpClient/Swagger or something else that is acceptable.

frantuma:
@fehguy yes for the given url and possibly similar. Renamed to Apache-HttpClient/Swagger

Issue #146, opened 2015-12-07
Using YAML anchors causes com.fasterxml.jackson.databind.node exception


chandravadans:
YAML anchors don't seem to be supported in swagger parser. An exception relating to conversion of `com.fasterxml.jackson.databind.node.TextNode`  to `com.fasterxml.jackson.databind.node.ArrayNode` occurs.

Example YAML :

``` yaml
swagger: '2.0'

info:
  version: "0.0.1"
  title: API

x-types:
  OperationType: &OperationType
    - registration

# Describe your paths here
paths:
  /checker:
    get:
      parameters:
        - name: operations
          in: query
          type: array
          items:
            type: string
            enum: *OperationType
          default: [registration]
      responses:
        200:
          description: OK
          schema:
            $ref: '#/definitions/OperationType'

definitions:      
  OperationType:
    type: string
    enum: *OperationType
```

example taken from [here](https://github.com/swagger-api/swagger-editor/issues/609) and the same issue filed [here](https://github.com/swagger-api/swagger-codegen/issues/1662)

nazr:
I'm having the same issue. That makes swagger-codegen pretty useless for any decent size swagger definition

ge1st:
I assume I have the same issue. The code generator CLI gives me the very helpful error:

```
[main] INFO io.swagger.codegen.DefaultCodegen - generated operationId puppiesGet       for Path: get /puppies
Exception in thread "main" java.lang.NullPointerException
        at io.swagger.codegen.DefaultGenerator.generateParameterId(DefaultGenerator.java:613)
        at io.swagger.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:531)
        at io.swagger.codegen.DefaultGenerator.processPaths(DefaultGenerator.java:496)
        at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:226)
        at io.swagger.codegen.cmd.Generate.run(Generate.java:188)
        at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:35)
```

Using one of the debug switches to the gen cli showed that many objects within the spec had null values in the object parsed in. Those with nulls were using anchors. Piping it in with a ruby yaml parser and back out as json let me work-around this.

I agree with @nazr this makes it difficult to handle large APIs.

fehguy:
Yes, it does look like anchors are not supported in the jackson parser that we use.  It may be supported in a later version but it is not expected to work right now.

[...]

PR #173, opened 2016-02-01
Refs #146 - Yaml anchor refs parsing. Direct usage of snakeyaml

frantuma:
@fehguy refs #146, also related to https://github.com/swagger-api/swagger-codegen/issues/1662 and https://github.com/swagger-api/swagger-editor/issues/609

This solution introduces "direct" usage of snakeyaml, which handles correctly scenarios as in #146  example.

It updates YAML "data as string" reading, replacing jackson-dataformat-yaml custom streaming solution with a snakeyaml load of Object, then converted/read by Jackson into JSON, probably at the expenses of performance.


fehguy:
Awesome--can you see about getting a tests in with a YAML anchor for both local (file system) and remote (url) cases?

frantuma:
@fehguy Added test for file as in example of #146. I am not sure what you mean in your comment, I guess YAML anchors only handle internal references,  possibly you refer to swagger `$ref` handling? Plsease let me know, in this case I'd say it would belong to separate ticket?

Issue #189, opened 2016-02-09
Parsing Swagger Doc with Path Parameter Not in Path


kevinmeredith:
Given `res3` equals the following simple Swagger Doc.

_Note_ that, per this [StackOverflow answer on path Swagger Path Params](http://stackoverflow.com/questions/27155516/swaggerissue-with-path-parameter/27156661#27156661), I did not append the path, `/{id}`, in the `/pets` path.

``` json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": "Swagger Petstore",
    "description": "A sample API that uses a petstore as an example to demonstrate features in the swagger-2.0 specification",
    "termsOfService": "http://swagger.io/terms/",
    "contact": {
      "name": "Swagger API Team"
    },
    "license": {
      "name": "MIT"
    }
  },
  "host": "petstore.swagger.io",
  "basePath": "/api",
  "schemes": [
    "http"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/pets": {
      "get": {
        "description": "Return a specific pet.",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "id of pet",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "a Pet."
          }
        }
      }
    }
  }
}
```

However, it parses successfully via `new Swagger20Parser().read(res3)`:

```
scala> import io.swagger.parser.Swagger20Parser
import io.swagger.parser.Swagger20Parser

scala> new Swagger20Parser().parse(res3)
res4: io.swagger.models.Swagger = io.swagger.models.Swagger@b6b5ef07

scala> res4.getPaths
res5: java.util.Map[String,io.swagger.models.Path] = {/pets=io.swagger.models.Path@23c163da}
```

Should there be a parse error since the `/pets` path did not equal `/pets/{id}` per the **path** parameter?

fehguy:
Yes, right now the parser is focused on structural aspects rather than content validation.  But I agree that it can be improved to handle cases like this.  I'll leave your ticket open for us to get it supported.

kevinmeredith:
Thanks for replying, @fehguy.

Has a discussion already happened on how to fix this? 

I'm only asking since I may have time to implement a fix. I'm looking at the code now, but, if you/others already have a fix in mind, that would help me out.

[...]

PR #196, opened 2016-02-11
Validate path parameters match parameters

kevinmeredith:
Addresses https://github.com/swagger-api/swagger-parser/issues/189.

Note - it's a **WIP**. I will `git rebase` when I believe it's ready for review. I'm using the PR to review the work that I've done via the PR "Files Changed" UI.

The intention of this PR is to fail, i.e. return `null` for the parsed `Swagger` object, if any `path`'s value, e.g. `/pets/{id}`, has an unaccounted for path parameter. Note that, for a path parameter of `{id}`, the global, path, and operation-specific path parameters would be retrieved, and then checked whether they, collectively, contain `{id}`.

I've added quite of bit of tests, but I'm going to add more to account for path parameters with global and path-specific path parameters.

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/Swagger20Parser.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/Swagger20Parser.java
@@ -2,6 +2,9 @@ package io.swagger.parser;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.BooleanNode;
+import com.google.common.base.Optional;
+import com.sun.org.apache.xpath.internal.operations.Bool;
 import io.swagger.models.Swagger;
 import io.swagger.models.auth.AuthorizationValue;
 import io.swagger.parser.util.ClasspathHelper;
@@ -9,6 +12,7 @@ import io.swagger.parser.util.DeserializationUtils;
 import io.swagger.parser.util.RemoteUrl;
 import io.swagger.parser.util.SwaggerDeserializationResult;
 import io.swagger.parser.util.SwaggerDeserializer;
+import static io.swagger.parser.util.SwaggerDeserializer.allPathParametersAccountForExtractedPathParameters;
 import io.swagger.util.Json;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.Validate;
@@ -129,6 +133,23 @@ public class Swagger20Parser implements SwaggerParserExtension {
                 SwaggerDeserializationResult result = new SwaggerDeserializer().deserialize(rootNode);
 
                 Swagger convertValue = result.getSwagger();
+
+                if(convertValue.getPaths() == null) {
+                    // no need to validate Path Parameters since there are no paths
+                }
+                else {
+                    final Optional<Boolean> validPaths = allPathParametersAccountForExtractedPathParameters(convertValue);
+                    if(validPaths.isPresent()) {
+                        final Boolean validResult = validPaths.get();
+                        if(validResult.equals(Boolean.FALSE)) {
+                            return null;
+                        }
+                    }
+                    else {
+                        return null;
+                    }
+                }
+
                 if (System.getProperty("debugParser") != null) {
                     LOGGER.info("\n\nSwagger Tree convertValue : \n"
                         + ReflectionToStringBuilder.toString(convertValue, ToStringStyle.MULTI_LINE_STYLE) + "\n\n");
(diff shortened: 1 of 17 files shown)


kevinmeredith:
> I've added quite of bit of tests, but I'm going to add more to account for path parameters with global and path-specific path parameters.

Done.

Reviewers, please check out this PR.

kevinmeredith:
@webron pointed out that my PR is wrong to include top-level `parameters`: https://github.com/swagger-api/swagger-parser/issues/189#issuecomment-183083193.

When I've updated the PR accordingly per Ron's reply, I will post here.

kevinmeredith:
I just updated this PR to exclude `global` parameters, per Ron's comment.

However, a single test failed:

```
Failed tests: 
  SwaggerReaderTest.testIssueSwaggerUi1792:114 NullPointer
```

This test uses `thingy.json`, whose ` `"/things/{id}"` path has an unaccounted for `id` path parameter. In other words, the `id` path parameter is not specified in either the path or operation `parameters`.

Should I just add `id` as either type of parameter, and then re-run the test to verify it succeeds?

fehguy:
Let's leave your test--I will verify a fix today and put it in when merging.  Thanks for the contrib!

kevinmeredith:
> Let's leave your test

To confirm that I understand - does this mean that I should leave the PR as-is, i.e. you'll verify the fix and fix the broken test, @fehguy?

Thank you

fehguy:
That's correct--I'll review the broken test and make sure it's either passing with another fix, or the test is disabled if it doesn't make sense.

kevinmeredith:
Hi @fehguy - could you please let me know if you've had a chance to review this PR?

Thanks

kevinmeredith:
Any updates, please? 

Thanks

fehguy:
Sorry for the delay.  I've been traveling and this takes a bit of time--I will review today.

modules/swagger-parser/src/main/java/io/swagger/parser/Swagger20Parser.java
```diff
@@ -129,6 +133,23 @@ private Swagger convertToSwagger(String data) throws IOException {
...
+
+                if(convertValue.getPaths() == null) {
+                    // no need to validate Path Parameters since there are no paths
+                }
+                else {
+                    final Optional<Boolean> validPaths = allPathParametersAccountForExtractedPathParameters(convertValue);
+                    if(validPaths.isPresent()) {
+                        final Boolean validResult = validPaths.get();
+                        if(validResult.equals(Boolean.FALSE)) {
+                            return null;
```
fehguy: We can't abort the parse if there are parameters missing.  The `SwaggerDeserializer` is responsible for keeping track of messages and returning a potentially invalid spec if there are problems.

modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
```diff
@@ -174,6 +176,176 @@ else if(!ROOT_KEYS.contains(key)) {
         return output;
     }
 
+    public static Optional<Boolean> allPathParametersAccountForExtractedPathParameters(final Swagger swagger) {
```
fehguy: Why use `Optional<Boolean>` instead of `boolean`?  We only have true/false for this operation.

fehguy:
OK @kevinmeredith I've gone though this and there are some challenges.

First, conceptually the parser's job is to _parse_ the spec, and provide messages back if it fails.

Next, the parser should not return null unless it's unreadable, or of a different version than the parser expects.  Null means "I don't know what to do" which is different from "this is a swagger 2.0 definition but there are issues".

Finally, and most importantly, we _cannot_ validate parameters unless we either (a) do full resolution in the validation process (such as, resolving remote references) or (b) check structural validation issues _after_ we have run the Resolver.

That said, there are some valuable checks in here, but I don't see how we can use the PR without some major changes.

I suggest working in the `SwaggerParser` and adding logic to this method:

```
SwaggerDeserializationResult readWithInfo(String location, List<AuthorizationValue> auths, boolean resolve)
```

_If_ `resolve = true`, you can then check for existing parameters _then_ add them to the messages in the `SwaggerDeserializationResult` as feedback rather than an error.

kevinmeredith:
Thanks for the feedback. I expect to reply to it, i.e. ask questions about particular changes, within 1 week or so.

kevinmeredith:
Sorry for the delay in addressing your comments, Tony. Thanks for your time in reviewing it.

In the event that I get time to issue another PR, I will reference this one.

Thanks

fehguy:
no worries, I'm trying to get a release out--we can always reopen or start a new ticket when it makes sense.

PR #232, opened 2016-04-11
update interfaces 

### LINKED ISSUE #231: parser interfaces need cleanup
There are too many ways to invoke the parser with confusing names and return types.
- In general, want to return the `SwaggerDeserializationResult` which has parse messages
- Want to throw an exception on invalid parse rather than returning null
- Names are all janky.  Fix them

Note this will be a breaking API change to the parser.

fehguy:
Fixes #231

### DIFF
--- a/modules/swagger-compat-spec-parser/pom.xml
+++ b/modules/swagger-compat-spec-parser/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>io.swagger</groupId>
         <artifactId>swagger-parser-project</artifactId>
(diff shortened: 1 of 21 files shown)


diegode:
@fehguy this seems great. Is there anything missing (besides the merge conflicts)?

fehguy:
I think we'll get the merge conflicts handled, but I did want to get another set of eyes on the proposed changes.  If you have any input, now would be a good time.

fehguy:
I'm going to close this out, since it has drifted a bunch.  Will revisit interfaces with 3.0 spec support.

PR #246, opened 2016-05-19
Don't configure the parent with allOf

### LINKED ISSUE #245: With a composed model (allOf), only set the model which holds a discriminator field as parent
As per swagger spec's definition of [allOf](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism), inheritance should only be used in the case the parent model has a discriminator field.
The current implementation is to set the first model as parent but this results in issues with generated code (see https://github.com/swagger-api/swagger-codegen/issues/2096)

cbornet:
Fix #245

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
@@ -813,7 +813,6 @@ public class SwaggerDeserializer {
             }
         } else if (allOf != null) {
             ComposedModel model = null;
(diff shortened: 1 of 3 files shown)


jeff9finger:
+1

cbornet:
rebased to resolve conflict.

cbornet:
@fehguy Thanks for merging :+1: !
Any chance to see a release soon ? This would allow to fix the Java inheritance issues in swagger-codegen !

fehguy:
@cbornet very close, this is required for the codegen release, which we (and you!) are working on.

PR #251, opened 2016-05-31
Generate better error messages

pablograna:
SwaggerDeserializationResult now contains the errors generated by
jackson if json parsing fails.

### DIFF
--- a/modules/swagger-compat-spec-parser/pom.xml
+++ b/modules/swagger-compat-spec-parser/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>io.swagger</groupId>
         <artifactId>swagger-parser-project</artifactId>
-        <version>1.0.20-SNAPSHOT</version>
+        <version>1.1.0-SNAPSHOT</version>
         <relativePath>../..</relativePath>
     </parent>
     <modelVersion>4.0.0</modelVersion>
--- a/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
+++ b/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
@@ -21,7 +21,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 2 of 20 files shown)


fehguy:
Thank you for this.  I'm not sure if the end user wants to see stack traces in the human-readable message, but it could be helpful in troubleshooting the problems.

One request though, please rebase to the develop branch, which will be part of 1.1.0.  There are some signature changes to clean up some ugly evolutions in how the parser is used.  We won't be releasing any new functionality in the 1.0.x series of the parser.

pablograna:
I am not 'dumping' the stack trace in the SwaggerDeserializationResult messages, I am just adding the error message as returned from Jackson. If there is a syntactic error in the json input, the result of the messages will be something like:

0: Unexpected character in ....
1: Line 1, column 1

This will let a tool (for example, swagger-codegen) provide more detail output to diagnose the problem, without dumping a stack trace.

I see that there are lot of changes in develop, I'll try to apply mine there and update the PR.

Thanks for the feedback.

fehguy:
@pablograna sorry if that came across as negative, I didn't mean it to be so.

pablograna:
One question, after a quick look at the develop branch: all parsing operations in SwaggerParser now return a SwaggerDeserializationResult instance. But they also throw the UnparseableContentException, so there are two ways of reporting errors. The only place I could see that a new UnparseableContentException is thrown are in SwaggerCompatConverter (looks like the methods throwing that exception are not implemented) and in AbstractParser.stringToNode. The other just sort of propagate an UnparseableContentException thrown by the Swagger20Parser or an extension.

Long story short: can the parser just report errors using the SwaggerDeserializationResult and avoid the UnparseableContentException altogether?

katari:
@fehguy  I did not consider your comment as negative, I apologize if my reply seemed that way (maybe the quotes around dump?). It was not at all my intention. I am very used to receiving critics to my code, (we do code reviews at work), so I take no offense at all .

Please, again, don't take my answer as being offended, it was just a 'terse' (is that the right term?) explanation.

I do really appreciate any feedback!!!!

Thanks for taking the time, and here goes another apology :)

pablograna:
@fehguy I was logged in as the wrong user in github!!!! This katari guy is me :)

sorry for the confusion.

pablograna:
Closing, need to change base to develop.

PR #252, opened 2016-06-01
Generate better error messages

pablograna:
SwaggerDeserializationResult now contains the errors generated by jackson if json parsing fails.

There is a disabled test that considers the case when the spec is provide as a string. I'll add a different commit that will change the api of AbstractParser.stringToNode (will throw a Jackson related exception).


pablograna:
AH!!!! NICE!!! Thanks a lot.

PR #254, opened 2016-06-03
Simplified extension interface

pablograna:
Made the SwaggerParserExtension much simpler: just one operation. The other operations are implemented in SwaggerParser and end up delegating in  SwaggerDeserializationResult.parseContents(... ).

This change removed some duplicate code, and makes the code easier to maintain (my opinion, of course).

Any opinions?

Another thing I'd like to do is to add another operation to SwaggerDeserializationResult: boolean supports(JsonNode swaggerAsJson). This will return true if the extension can parse the swagger spec. The two extensions I know about are the Swagger20Parser and SwaggerCompatConverter. In both these extensions it is very easy to check if they support the spec.

Thanks a lot for your hard work in swagger.


pablograna:
I added a second commit that changes the way SwaggerParser calls extensions. Before, SwaggerParser first called Swagger20Parser, then it tried with all extensions. I added a supports() operation to the SwaggerParserExtension interface and SwaggerParser asks each extension if it supports the swagger spec. It then delegates the parsing to that extension.

Please, let me know if this makes sense ...

fehguy:
I know there are reasons for using the URL or sting contents. Let me jog the brain first and bring them up for discussion.

pablograna:
@fehguy Thanks for your comment. Let me know what you think and if you'd like me to change anything.

One note, though: if clients use SwaggerParser, they will sill be able to pass a URL or a string directly. But if the intention is to expose the extensions to clients, then yes, you are right, and I should keep the interface unchanged.

Thanks a lot!

pablograna:
A follow up note: I have a local swagger-codegen local branch updated to use this version of the parser. If anyone is interested, I could upload the diff.

pablograna:
@fehguy, did you have a change to look at this? What do you think?

fehguy:
Hi, I've been slammed, so apologies for the delay.

So here's the challenge.  The parser is responsible for reading a swagger definition and converting it to a swagger object.  It shouldn't actually care if it's using Jackson or not, and the problem with Jackson + swagger definitions is that there is no way to read YAML files in a lossless fashion, as comments are lost in the process.

You could say "well there are no comments in the swagger POJOs"--which is sort of true, one can easily use the extensions mechanism to store comments in the POJO--but by changing the interface to the parser to _require_ a `JsonNode`, we are effectively making the ability to handle YAML correctly impossible.

The other issue is something that you've had to do in your tests--that is, calling an HTTP library to load the specification on your own, when the Parser itself is quite capable of doing so, and is required to do so because it may need to resolve external references.

So my suggestion is to continue to support the following inputs into the parser for reading & resolving into POJOS:

* Swagger location via URL.  Currently this is a string, but it could be better done as a java.net.URL
* Swagger definition via String.  This would allow parsing YAML and gives the opportunity to handle the YAML more elegantly instead of jamming in a lossy fashion into a JsonNode object (I have a lossless YAML parser almost done)
* Swagger object via `JsonNode`

I do agree the signatures could be improved though

pablograna:
Thanks @fehguy. Good point about the docs in yaml. I'd like to know a little bit more about the lossless YAML parser (btw: jackson may also parse json comments, a non-standard feature, but I think comments are not available from JsonNode). I still think that the parsing interface is too big. In my PR I just disregarded the complexity of resolving external references, just left everything as it was originally.

fehguy:
I'm going to close this out for now, the 1.1.x version will have a better interface, and we can have a plan for YAML with a parser that we've been working on that preserves comments.

Issue #271, opened 2016-07-22
Path parameters are being duplicated in certain cases


diegode:
This test is failing in master:

    @Test
    public void testUniqueParameters() throws Exception {
        String yaml =
                "swagger: '2.0'\n" +
                "info:\n" +
                "  title: test\n" +
                "  version: '0.0.0'\n" +
                "parameters:\n" +
                "  foo-id:\n" +
                "    name: id\n" +
                "    in: path\n" +
                "    type: string\n" +
                "    required: true\n" +
                "paths:\n" +
                "  /foos/{id}:\n" +
                "    parameters:\n" +
                "        - $ref: '#/parameters/foo-id'\n" +
                "    get:\n" +
                "      responses:\n" +
                "        200:\n" +
                "          schema:\n" +
                "            $ref: '#/definitions/foo'\n" +
                "    put:\n" +
                "      parameters:\n" +
                "        - name: foo\n" +
                "          in: body\n" +
                "          required: true\n" +
                "          schema:\n" +
                "            $ref: '#/definitions/foo'\n" +
                "      responses:\n" +
                "        200:\n" +
                "          schema:\n" +
                "            $ref: '#/definitions/foo'\n" +
                "definitions:\n" +
                "  foo:\n" +
                "    type: object\n" +
                "    properties:\n" +
                "      id:\n" +
                "        type: string\n" +
                "    required:\n" +
                "      - id\n";
        SwaggerParser parser = new SwaggerParser();
        Swagger swagger = parser.parse(yaml);
        List<Parameter> parameters = swagger.getPath("/foos/{id}").getPut().getParameters();
        assertTrue(parameters.size() == 2);
    }

diegode:
@fehguy you can close this :)

PR #272, opened 2016-07-22
Fixes issue 271

diegode:
This is for https://github.com/swagger-api/swagger-parser/issues/271


diegode:
@fehguy I think this is safe to merge

Issue #270, opened 2016-07-20
Nested refs aren't resolved


sartymlg:
Hi, In my definition I have a lot of schemas, so I decided split them in multiples files, the problem arises when in an external file, there are another reference.

As example, I've split this [file](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/yaml/petstore.yaml
) into this [gist](https://gist.github.com/sartymlg/4ea56197ea412e002ff3dc1307ed9339) 

After parse pets.yaml, only there are two definitions:
- Pets
- Error
However Pet is not added into definition field, so codegen can't generate it. 

I think there are two problems:
- Resolve function isn't recursive
- When a sub schema is loaded from a external file, actual genericRef is INTERNAL instead RELATIVE

PS: This issue is probably related with #229 and #147

diegode:
@sartymlg can you check if my PR fixes your issue?

sartymlg:
I've resolved it in this branch https://github.com/TheWorkshopLabs/swagger-parser/commits/SD-133-code-generation-error-external-refs

diegode:
@sartymlg can you create a pull request? I think we both did complementary things.

diegode:
@fehguy perhaps this should be reopened, since there's still work to do (perhaps add https://github.com/swagger-api/swagger-parser/compare/master...TheWorkshopLabs:SD-133-code-generation-error-external-refs).

fehguy:
@diegode I can create a PR, please take a look when you can

diegode:
@fehguy you can close this :)

PR #273, opened 2016-07-26
Fixes issue 270

diegode:
This is for https://github.com/swagger-api/swagger-parser/issues/270


diegode:
@fehguy I think this is safe to merge

Issue #278, opened 2016-08-02
External recursive refs aren't being resolved


diegode:
This test is failing in master (the last assertion):

    @Test
    public void testLoadRecursiveExternalDef() throws Exception {
        SwaggerParser parser = new SwaggerParser();
        final Swagger swagger = parser.read("src/test/resources/file-reference-to-recursive-defs/b.yaml");
        Map<String, Model> definitions = swagger.getDefinitions();
        assertEquals(((RefProperty) ((ArrayProperty) definitions.get("v").getProperties().get("children")).getItems()).get$ref(), "#/definitions/v");
        assertTrue(!definitions.containsKey("y"));
        assertEquals(((RefProperty) ((ArrayProperty) definitions.get("x").getProperties().get("children")).getItems()).get$ref(), "#/definitions/x");
    }

where a.yaml is:

    ---
    swagger: "2.0"
    info:
      title: test
      version: '0.0.0'
    paths: {}
    definitions:
      y:
        type: object
        properties:
          name:
            type: string
          children:
            type: array
            items:
              $ref: "#/definitions/y"

and b.yaml is:

    ---
    swagger: "2.0"
    info:
      title: test
      version: '0.0.0'
    paths: {}
    definitions:
      x:
        $ref: "./a.yaml#/definitions/y"
      v:
        type: object
        properties:
          name:
            type: string
          children:
            type: array
            items:
              $ref: "#/definitions/v"

fehguy:
This has been fixed

diegode:
@fehguy please re-enable the test from my PR (https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java#L226)

PR #279, opened 2016-08-02
Adds failing test for issue278

diegode:
This is for https://github.com/swagger-api/swagger-parser/issues/278


fehguy:
merged with test disabled for now

PR #280, opened 2016-08-03
Sd 133 code generation error external refs

fehguy:
see https://github.com/swagger-api/swagger-parser/pull/273

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
@@ -1,5 +1,6 @@
 package io.swagger.parser.processors;
 
+import io.swagger.models.ArrayModel;
 import io.swagger.models.Model;
 import io.swagger.models.Swagger;
 import io.swagger.models.properties.Property;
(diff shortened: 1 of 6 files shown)


modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
```diff
@@ -59,6 +60,10 @@ public String processRefToExternalDefinition(String $ref, RefFormat refFormat) {
                         subRef.set$ref(processRefToExternalDefinition(subRef.get$ref(), subRef.getRefFormat()));
                 }
             }
+        } else if (model instanceof ArrayModel && ((ArrayModel)model).getItems() instanceof RefProperty) {
```
diegode: Perhaps this shouldn't be an "else", but a separate "if" below.

modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
```diff
@@ -59,6 +60,10 @@ public String processRefToExternalDefinition(String $ref, RefFormat refFormat) {
...
+            RefProperty subRef = (RefProperty) ((ArrayModel)model).getItems();
+            if(isAnExternalRefFormat(subRef.getRefFormat()))
+                subRef.set$ref(processRefToExternalDefinition(subRef.get$ref(), subRef.getRefFormat()));
```
diegode: I think here we need to add an "else" case too, like here: https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java#L71
(perhaps you should rebase this branch on top of master)

diegode:
Shouldn't we also consider the case of ComposedModel? (and I'm not sure if there are others).

diegode:
@fehguy I just created this other pull request: https://github.com/swagger-api/swagger-parser/pull/284
that fixes a different but related case.

diegode:
@fehguy you can close this case and merge instead https://github.com/swagger-api/swagger-parser/pull/285

PR #300, opened 2016-09-14
bug with readOnly and vendor extensions in properties

fiore-adrian:


fiore-adrian:
issue:
https://github.com/swagger-api/swagger-parser/issues/301

fehguy:
`readOnly` added in #364, the vendor extensions support you added isn't legal in the spec :(

PR #303, opened 2016-09-28
Parse external references inside vendor extensions

diegode:
This adds support for parsing external references inside vendor extensions, i.e: whenever a $ref is found inside a vendor extension that points to another file, the referenced definition/parameter/response is added to the swagger object.

I'm not sure if this is the intended behavior by the Swagger spec, but it is useful for adding complex behavior in vendor extensions without forking the swagger-parser.


fehguy:
Hi, we don't support this in the spec, so have to punt in the parser :(  Consider adding an extension to do this, and I'd be happy to add it to the repo as well.

Issue #319, opened 2016-10-26
Second level of ref object not parsing correctly


gustavorodrig:
Swagger parser doesn't parse correctly many levels of references when in the second reference are an object that is referencing another file

Imagine this scenario:

`
Main.yaml:

Mail:
    type: object
    properties:
      message:
          $ref: "../files/Definitions.yaml#/definitions/Message"

-------------------------
Definitions.yaml:
    
Message:
    type: object
    properties:
      content:
        type: object
        properties:
         contentMessage:
           $ref: "../files/others.yaml#/Content" 

-----------------------------
Others.yaml:
    
Content:
    type: object
    properties:
      text:
        type: string
`

In this scenario swagger creates a POJO (ContentMessage) withou any properties and dont create the last object (Content)

I added some code to consider this situation and i will do a pull request.

Issue #318, opened 2016-10-26
Support *.yml files as yaml format


gustavorodrig:
Swagger parser is only considering *.yaml files when parsing, for some reason many companies uses *.yml in their swagger projects..

I modified the code to consider *.yml as YAML format and I will do a pull request to project.

fehguy:
This is a duplicate of #317

PR #320, opened 2016-10-26
Issue #318 and #319

gustavorodrig:


fehguy:
See my comment, also looks like you have a build or test error

fehguy:
Addressed by #322

PR #333, opened 2016-10-27
added check to avoid removing referenced models #278

### LINKED ISSUE #278: External recursive refs aren't being resolved

fehguy:
Fixes #278 

@diegode this is to fix the issue in #278 but...

I don't necessarily agree with the expectations in a couple tests.  Once I fixed this bug some other test failures appeared.

For example, when we have this:

```yaml
definitions:
  x:
    properties:
      foo:
        $ref: '#/definitions/y'
  y:
    $ref: '#/definitions/z'
  z:
    type: object
```

That the parser should change the definitions to this:

```yaml
definitions:
  x:
    properties:
      foo:
        $ref: '#/definitions/z'
  z:
    type: object
```

Because a model is referencing `y` in this example as a property, there may be intent and _value_ in retaining that name.  So even though `y` absorbs all the properties of `z`, the name `y` may be useful.

So this PR changes the behavior of a couple tests that were previously behaving differently.  Since you put them there, I thought it was only considerate to get your feedback before any merge :)

### DIFF
--- a/modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
+++ b/modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
@@ -185,9 +185,10 @@ public class SwaggerParserTest {
     public void testLoadExternalNestedDefinitions() throws Exception {
         SwaggerParser parser = new SwaggerParser();
         final Swagger swagger = parser.read("src/test/resources/nested-references/b.yaml");
+
         Map<String, Model> definitions = swagger.getDefinitions();
         assertTrue(definitions.containsKey("x"));
-        assertTrue(!definitions.containsKey("y"));
+        assertTrue(definitions.containsKey("y"));
         assertTrue(definitions.containsKey("z"));
         assertEquals(((RefModel) definitions.get("i")).get$ref(), "#/definitions/k");
     }
@@ -223,15 +224,15 @@ public class SwaggerParserTest {
         assertNotNull(Yaml.mapper().writeValueAsString(swagger));
     }
 
-    @Test(enabled = false)
+    @Test
     public void testLoadRecursiveExternalDef() throws Exception {
         SwaggerParser parser = new SwaggerParser();
         final Swagger swagger = parser.read("src/test/resources/file-reference-to-recursive-defs/b.yaml");
 
         Map<String, Model> definitions = swagger.getDefinitions();
         assertEquals(((RefProperty) ((ArrayProperty) definitions.get("v").getProperties().get("children")).getItems()).get$ref(), "#/definitions/v");
-        assertTrue(!definitions.containsKey("y"));
-        assertEquals(((RefProperty) ((ArrayProperty) definitions.get("x").getProperties().get("children")).getItems()).get$ref(), "#/definitions/x");
+        assertTrue(definitions.containsKey("y"));
+        assertEquals(((RefProperty) ((ArrayProperty) definitions.get("x").getProperties().get("children")).getItems()).get$ref(), "#/definitions/y");
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -185,9 +185,10 @@ public void testLoadRelativeFileTree_Json() throws Exception {
     public void testLoadExternalNestedDefinitions() throws Exception {
         SwaggerParser parser = new SwaggerParser();
         final Swagger swagger = parser.read("src/test/resources/nested-references/b.yaml");
+
         Map<String, Model> definitions = swagger.getDefinitions();
         assertTrue(definitions.containsKey("x"));
-        assertTrue(!definitions.containsKey("y"));
```
fehguy: Note that `y` lives on, even though it is directly referencing all aspects of `z`

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -223,15 +224,15 @@ public void testLoadRelativeFileTree_Yaml() throws Exception {
...
     public void testLoadRecursiveExternalDef() throws Exception {
         SwaggerParser parser = new SwaggerParser();
         final Swagger swagger = parser.read("src/test/resources/file-reference-to-recursive-defs/b.yaml");
 
         Map<String, Model> definitions = swagger.getDefinitions();
         assertEquals(((RefProperty) ((ArrayProperty) definitions.get("v").getProperties().get("children")).getItems()).get$ref(), "#/definitions/v");
-        assertTrue(!definitions.containsKey("y"));
-        assertEquals(((RefProperty) ((ArrayProperty) definitions.get("x").getProperties().get("children")).getItems()).get$ref(), "#/definitions/x");
+        assertTrue(definitions.containsKey("y"));
+        assertEquals(((RefProperty) ((ArrayProperty) definitions.get("x").getProperties().get("children")).getItems()).get$ref(), "#/definitions/y");
```
fehguy: Instead of changing the reference to `x`, I'm leaving it as modeled as `y`

fehguy:
OK I've reviewed the behavior with a few others and this PR and test changes seem correct.  We don't want this:

```
definitions:
  Address:
    type: object
    # properties:...
  AlternateAddress:
    $ref: '#/definitions/Address'
```

to remove the `Address` definition and change references to `AlternateAddress`.  @diegode please open a new issue if you disagree.

PR #354, opened 2016-12-05
Fix bug in relative path generation: 

samanthacatania:
'./' results in array out of bounds ands '../' results in root component not being truncated properly


fehguy:
Can you please add a test to reproduce this issue?

samanthacatania:
I added tests and simplified the code a bit. However I found an issue with nested relative paths. I'' try to figure out how to fix it.

samanthacatania:
Opened the following issue #355

fehguy:
Please use master.  The develop branch has drifted a bunch and will be deleted.

PR #374, opened 2016-12-27
Support file URL scheme produced by java.io.File et al.

kevinoid:
The URLs produced by `java.io.File.toURI()` (and deprecated `.toURL()`) and `java.lang.Class.getResource(String)` contain no additional slashes between `file:` and the path.  This makes using `Swagger20Parser` with files from these methods a bit more painful than it needs to be.  This PR addresses this by recognizing strings starting with `file:` as file URLs and adds tests to cover this case.

I hope you find it useful,
Kevin

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/Swagger20Parser.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/Swagger20Parser.java
@@ -42,7 +42,7 @@ public class Swagger20Parser implements SwaggerParserExtension {
             if (location.toLowerCase().startsWith("http")) {
                 data = RemoteUrl.urlToString(location, auths);
             } else {
-                final String fileScheme = "file://";
+                final String fileScheme = "file:";
                 Path path;
                 if (location.toLowerCase().startsWith(fileScheme)) {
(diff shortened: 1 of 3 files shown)


fehguy:
Thanks @kevinoid.  Can you verify this fix works as expected on Windows?

kevinoid:
@fehguy I don't have a Windows Java development environment immediately available, but I could setup [AppVeyor](https://www.appveyor.com/) pretty quickly.  Would you like a PR to add an `appveyor.yml` or would you prefer I run a one-off test?

kevinoid:
@fehguy I made a branch with [appveyor.yml](https://github.com/kevinoid/swagger-parser/blob/appveyor/appveyor.yml).  I can send a PR for that if you would like.

Testing this PR revealed a spot which I had overlooked in `io.swagger.parser.util.PathUtils` which I corrected so it now [passes](https://ci.appveyor.com/project/kevinoid/swagger-parser/build/1.0.9)

fehguy:
@kevinoid sure, please send a PR for the `appveyor.yml` and we'll make that part of the regular testing process.  Thanks!

Issue #398, opened 2017-02-21
Issues encountered with swagger-parser


quilicicf:
### Issues

We tried to deserialize a variety of Swagger definitions with the parser and encountered the following issues:

|Issue|More information|
|---|---|
|API key security schemes description are lost | It is simply a missing line in the code |
|Security definitions order is not kept | The implementation of Map is HashMap, should be LinkedHasmap (even though JSON is not supposed to keep map entries order, it's a nice to have for code versioning) |
|Enums on non-string parameters are lost | The values are added via JsonNode#textValue without type check (only works on TextNode)|
|Empty schemas are lost | There is specific code to avoid that, we use it as a fallback for things we don't handle yet so in our specific use-case it is an issue |

I can make specific tickets/implement fixes/discuss any of these issues.

### How to reproduce

Below are a Swagger and Java test file. I'd expect all the tests in it to pass.

#### Swagger file

```yaml
---
swagger: "2.0"
info:
  version: 1.0.0
  title: Test file

paths: {}

parameters:
  invalid:
    name: invalid
    in: header
    type: integer
    format: int32
    enum:
      - 12

securityDefinitions:
  "Basic auth scheme":
    type: basic
  "API Key scheme":
    type: apiKey
    description: Lost in translation
    name: auth
    in: header

definitions:
  EmptySchema: {}

```

#### Java file

```java
import com.google.common.collect.Iterables;
import io.swagger.models.Swagger;
import io.swagger.models.auth.SecuritySchemeDefinition;
import io.swagger.models.parameters.AbstractSerializableParameter;
import io.swagger.parser.SwaggerParser;
import io.swagger.parser.util.SwaggerDeserializationResult;
import org.junit.Assert;
import org.junit.Test;

import java.io.IOException;
import java.nio.file.Paths;
import java.util.Map;

public class TestClass {

    @Test
    public void testSwaggerParserIssues() throws IOException {
        final SwaggerParser swaggerParser = new SwaggerParser();

        // TODO: Update according to your environement
        final String swaggerPath = "/path/to/provided/file.swagger";
        final SwaggerDeserializationResult parsingResult = swaggerParser.readWithInfo(swaggerPath, null, true);

        final Swagger swagger = parsingResult.getSwagger();

        // Security definitions order issue
        final Map.Entry<String, SecuritySchemeDefinition> firstSecurityScheme =
                Iterables.get(swagger.getSecurityDefinitions().entrySet(), 0);
[...]

PR #404, opened 2017-02-28
Fix/misc

boillodmanuel:
Several fixes related to https://github.com/swagger-api/swagger-parser/issues/398.

I did one commit per fix (see commit message) to facilitate the review

I'll be happy to discuss each commit
- commit `#e33949d` with message `feat(resolve): add settings to allow resolving references without adding resource parameters to each operation` do not change public behaviour (of `SwaggerParser`) but for those using directly `SwaggerResolver`, it allow to don't add resource parameters on operation, and have a behaviour closer to `ObjectMapper` and also to keep the `PathParameter`on path (not flatten to each operation)

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
@@ -36,14 +36,14 @@ public class ResolverCache {
     private final List<AuthorizationValue> auths;
     private final Path parentDirectory;
     private final String rootPath;
-    private Map<String, Object> resolutionCache = new HashMap<>();
-    private Map<String, String> externalFileCache = new HashMap<>();
-    private Set<String> referencedModelKeys = new HashSet<>();
+    private Map<String, Object> resolutionCache = new LinkedHashMap<>();
+    private Map<String, String> externalFileCache = new LinkedHashMap<>();
+    private Set<String> referencedModelKeys = new LinkedHashSet<>();
 
     /*
     a map that stores original external references, and their associated renamed references
      */
-    private Map<String, String> renameCache = new HashMap<>();
+    private Map<String, String> renameCache = new LinkedHashMap<>();
 
     public ResolverCache(Swagger swagger, List<AuthorizationValue> auths, String parentFileLocation) {
         this.swagger = swagger;
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
@@ -14,7 +14,12 @@ import io.swagger.parser.util.SwaggerDeserializer;
 
 import java.io.File;
 import java.nio.file.Path;
-import java.util.*;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -36,14 +41,14 @@ public class ResolverCache {
     private final List<AuthorizationValue> auths;
     private final Path parentDirectory;
     private final String rootPath;
-    private Map<String, Object> resolutionCache = new LinkedHashMap<>();
-    private Map<String, String> externalFileCache = new LinkedHashMap<>();
-    private Set<String> referencedModelKeys = new LinkedHashSet<>();
+    private Map<String, Object> resolutionCache = new HashMap<>();
+    private Map<String, String> externalFileCache = new HashMap<>();
+    private Set<String> referencedModelKeys = new HashSet<>();
 
     /*
     a map that stores original external references, and their associated renamed references
      */
-    private Map<String, String> renameCache = new LinkedHashMap<>();
+    private Map<String, String> renameCache = new HashMap<>();
 
     public ResolverCache(Swagger swagger, List<AuthorizationValue> auths, String parentFileLocation) {
         this.swagger = swagger;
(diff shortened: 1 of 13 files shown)


modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
```diff
@@ -36,14 +36,14 @@
...
-    private Set<String> referencedModelKeys = new HashSet<>();
+    private Map<String, Object> resolutionCache = new LinkedHashMap<>();
+    private Map<String, String> externalFileCache = new LinkedHashMap<>();
+    private Set<String> referencedModelKeys = new LinkedHashSet<>();
 
     /*
     a map that stores original external references, and their associated renamed references
      */
-    private Map<String, String> renameCache = new HashMap<>();
+    private Map<String, String> renameCache = new LinkedHashMap<>();
```
fehguy: i don't think the `LinkedHashMap` is necessary here as order never matters in this cache.  It's just for resolution.

modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
```diff
@@ -36,14 +36,14 @@
     private final List<AuthorizationValue> auths;
     private final Path parentDirectory;
     private final String rootPath;
-    private Map<String, Object> resolutionCache = new HashMap<>();
-    private Map<String, String> externalFileCache = new HashMap<>();
-    private Set<String> referencedModelKeys = new HashSet<>();
+    private Map<String, Object> resolutionCache = new LinkedHashMap<>();
```
fehguy: HashMap should do fine, did you see an issue with it?

modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
```diff
@@ -36,14 +36,14 @@
...
+    private Map<String, String> externalFileCache = new LinkedHashMap<>();
```
fehguy: HashMap should do fine, did you see an issue with it? (repeating)

modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerResolver.java
```diff
@@ -53,4 +63,26 @@ public Swagger resolve() {
 
         return swagger;
     }
+
+    public static class Settings {
```
fehguy: It's not clear to me why we need this.  The need to add parameters to operations is dictated by the specification and the location of the parameters, not some user-specified setting.

modules/swagger-parser/src/main/java/io/swagger/parser/processors/DefinitionsProcessor.java
```diff
@@ -29,7 +29,7 @@ public void processDefinitions() {
             return;
         }
 
-        Set<String> keySet = new HashSet<>();
+        Set<String> keySet = new LinkedHashSet<>();
```
fehguy: As with before, not clear why we care about order for these internal storage mechanisms.

modules/swagger-parser/src/main/java/io/swagger/parser/processors/PathsProcessor.java
```diff
@@ -16,12 +17,17 @@
 
     private final Swagger swagger;
     private final ResolverCache cache;
+    private final SwaggerResolver.Settings settings;
```
fehguy: As with above, I don't see why we need to add a `Settings` here.

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerResolverTest.java
```diff
@@ -40,7 +41,7 @@ public void testSwaggerResolver(@Injectable final Swagger swagger,
             result = definitionsProcessor;
             times = 1;
 
-            new PathsProcessor(cache, swagger);
+            new PathsProcessor(cache, swagger, withInstanceOf(SwaggerResolver.Settings.class));
```
fehguy: I don't like this.  What else would be passed in the `PathsProcessor`?  Is this expected to be an extension point?

modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerResolver.java
boillodmanuel: The last 3 comments is about this `addParametersToEachOperation` setting.
I understand it as it is only useful is specific use cases.

First at all, the variable could have a better name `addResourceParametersToEachOperation`.
The current behaviour (before this PR) is:
- resource parameters are parsed
- when processing operations, resource parameters are merged with operation parameters at operation level and resources parameters are emptied.

This result to a `Swagger` object with no resource parameters.  Even `path variable` are cleaned from resource parameters, which is the kind of parameter which have the most value at resource level. 
In our usage of swagger-parser, we want to keep resource parameters as is.
So the best option I saw to add this capability without changing swagger-parser behaviour is to introduce this settings.

fehguy: I see.  So the parser is essentially copying the parameters from the path-item object to the operations "below" it in hierarchy correct?  and what you're saying is, you don't want it to copy them, or you want to copy and leave them there?

boillodmanuel: > So the parser is essentially copying the parameters from the path-item object to the operations "below" it in hierarchy correct

Exactly

> you don't want it to copy them, or you want to copy and leave them there?

I don't want to copy them at all.

Before, we were using the swagger-core Object-Mapper, but it didn't support multi-files. So we compare it with swagger-parser and I tried to fix all important difference with this PR.

PR #406, opened 2017-03-05
Example specification petstore.json is not compliant with the OpenAPI specification 2.0

obfischer:
This fixes a validation error of petstore.json against the OpenAPI specification 2.0. petstore.json will pass a validation against the OpenAPI specification 2.0.


fehguy:
Right you are!  Thanks

PR #408, opened 2017-03-08
Added support for $ref Headers

tompahoward:
We shouldn’t have to repeat the definitions for headers in every single
response.

Please don’t make me repeat the header definitions in every response
(ಥ﹏ಥ)

Please see modules/swagger-parser/src/test/resources/headers.yaml for
an example

@Helmsdown what’s your view? From what I can tell, you originally made
the comment that it wasn’t needed. If so, your feedback (positive or
negative) would be greatly appreciated.

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ResponseProcessor.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ResponseProcessor.java
@@ -1,30 +1,62 @@
 package io.swagger.parser.processors;
 
+import java.util.Map;
+
+import io.swagger.models.RefResponse;
 import io.swagger.models.Response;
 import io.swagger.models.Swagger;
 import io.swagger.models.properties.Property;
+import io.swagger.models.properties.RefProperty;
 import io.swagger.parser.ResolverCache;
 
 public class ResponseProcessor {
 
     private final PropertyProcessor propertyProcessor;
+    private final ResolverCache cache;
 
     public ResponseProcessor(ResolverCache cache, Swagger swagger) {
+        this.cache = cache;
         propertyProcessor = new PropertyProcessor(cache, swagger);
     }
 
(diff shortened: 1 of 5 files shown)


fehguy:
I hate to say this... and TBH I don't think it was intentional, but references are not supported in the header object (correct, @webron ?).  The spec is here:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#headers-object

So that means unfortunately the tool cannot support it.  We do our very best to not step outside the bounds of the spec to be as compatible as possible.

The good news is that the 3.0 spec, which supports this, hit RC0 the other day.  Which means support is on the way (that rhymes):

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#headerObject

webron:
Yup, unfortunately it is not supported in 2.0 😕

tompahoward:
👍 Fair call. I'd seen comments indicating this, but I thought I'd try my luck.

Help me out for a sec, if you can. What part of the spec specifies that references are not permitted in the header object? I've looked and I can't find it, but I might be looking in the wrong place.

For instance, I can see many examples where "items" contains a $ref, but the [Items Object specification](http://swagger.io/specification/#itemsObject) doesn't mention references at all.

In the [File Structure section](http://swagger.io/specification/#file-structure-11), it says 

> parts of the definitions can be split into separate files, at the discretion of the user

But it doesn't say much more than that.

I guess I'm asking how do people like myself, which are trying to implement conforming applications, figure out where $ref may or may not be used?

webron:
It's simple. Where it explicitly says it can be used.

You're right, the `Items Object` doesn't allow for $ref's. However, the Schema Object does. And the Schema Object _does not_ use the `Items Object` as the value for its `items. This subtle difference doesn't allow for query parameters to have arrays of models, for examples, but it does allow for array of models wherever a Schema Object.

Issue #411, opened 2017-03-09
Parser becomes sad, dies when remote reference is wrong type


fehguy:
It's user error, but the parser should be a bit more lenient.  If you have a `BodyParameter` and that has a schema reference which is invalid--either not reachable, not parseable, or is of the wrong type--the parser throws a fit and walks away without any warning.  While we shouldn't go too far out of the way in this case, we should at least not quit, and possibly send a nice message that there's an issue.

fehguy:
Fixed by #412

PR #412, opened 2017-03-09
better handling of remote references

### LINKED ISSUE #407: NPE on nested remote references
Oh snap.  Looks like this:

https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java#L128

```
result = (T) ser.definition((ObjectNode) tree, definitionPath.replace("/", "."), null);
```

Will pretty much always guarantee a NPE in the deserializer because it is expected that the `ParseResult` is never null :(

fehguy:
Fixes #407, #411

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java
@@ -102,7 +102,10 @@ public class SwaggerParser {
             if (result != null) {
                 result.setSwagger(new SwaggerResolver(result.getSwagger(), new ArrayList<AuthorizationValue>(), null).resolve());
             }
-            return new Swagger20Parser().readWithInfo(node);
+            else {
+                result.message("Definition does not appear to be a valid Swagger format");
+            }
+            return result;
         }
         catch (Exception e) {
             return new SwaggerDeserializationResult().message("malformed or unreadable swagger supplied");
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
@@ -34,7 +34,7 @@ public final class ExternalRefProcessor {
         final Model model = cache.loadRef($ref, refFormat, Model.class);
 
         if(model == null) {
-            // stop!  There's a problem
+            // stop!  There's a problem.  retain the original ref
             LOGGER.warn("unable to load model reference from `" + $ref + "`.  It may not be available " +
                     "or the reference isn't a valid model schema");
             return $ref;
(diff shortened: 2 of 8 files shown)


modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java
```diff
@@ -103,7 +103,10 @@ public SwaggerDeserializationResult readWithInfo(String swaggerAsString) {
             if (result != null) {
                 result.setSwagger(new SwaggerResolver(result.getSwagger(), new ArrayList<AuthorizationValue>(), null).resolve());
             }
-            return new Swagger20Parser().readWithInfo(node);
+            else {
+                result.message("Definition does not appear to be a valid Swagger format");
+            }
+            return result;
```
fehguy: I have no idea why we were parsing, resolving, then re-parsing without re-resolving.  Seems like a bad joke, but in the end, we were returning an un-resolved model!

modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
```diff
@@ -33,6 +33,12 @@ public ExternalRefProcessor(ResolverCache cache, Swagger swagger) {
     public String processRefToExternalDefinition(String $ref, RefFormat refFormat) {
         final Model model = cache.loadRef($ref, refFormat, Model.class);
 
+        if(model == null) {
+            // stop!  There's a problem
+            LOGGER.warn("unable to load model reference from `" + $ref + "`.  It may not be available " +
+                    "or the reference isn't a valid model schema");
+            return $ref;
```
fehguy: This is halfway to the goal, it doesn't die, but it's not nice in that there's no message sent back to the end user

modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
```diff
@@ -698,6 +698,10 @@ public Path pathRef(TextNode ref, String location, ParseResult result) {
     }
 
     public Model definition(ObjectNode node, String location, ParseResult result) {
+        if(result == null) {
+            // TODO, this shouldn't happen, but the `ResolverCache#loadRef` method is passing null
+            result = new ParseResult();
```
fehguy: I know!  `TODO` is very ugly.  But so is a NPE!  Will try to fix before merging

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -763,6 +763,27 @@ public void testIssue393() {
...
+                "info:\n" +
+                "  description: |\n" +
+                "  version: 1.0.0\n" +
+                "  title: testing\n" +
+                "paths:\n" +
+                "   /foo:\n" +
+                "     $ref: 'http://petstore.swagger.io/v2/swagger.json#/paths/~1pet'\n" +
+                "   /bar:\n" +
+                "     $ref: 'http://petstore.swagger.io/v2/swagger.json#/paths/~1pet'\n" +
+                "schemes:\n" +
```
fehguy: This relies on a run-time server being up for the test.  No bueno--will try to change this to use Mocks

PR #414, opened 2017-03-10
Issue 358 Path param matching fails

andylowry:
I've supplied a test that demonstrates the problem described by the
issue.

When the swagger model is parsed, the GET operation ends up with two identical parameters, because the override test in `PathsProcessor` was performed before the path parameter was resolved, causing the parameter not to match the operation parameter.


andylowry:
@fehguy I'm confused here - I don't see a commit that fixes the problem my test demonstrates. It looks like merging this PR has probably broken master. My PR did _not_ include a fix.

fehguy:
I merged your test and pushed a fix.  See https://github.com/swagger-api/swagger-parser/commit/0a6038220faeab15dbc65d7d9a56cf6003e6533e

andylowry:
OK, cool. Just being careful. Expected a fix would be part of this PR or another PR, and saw neither.

fehguy:
I can't update your fork so can't update your PR

andylowry:
Ah, makes sense. First time I've bumped up against that. Thanks

PR #422, opened 2017-03-13
Fix handling of relative refs without leading dot

### LINKED ISSUE #421: Relative refs in parameters and schema are not resolved correctly
If the value of a $ref is a relative reference but does not start with ".", the swagger core (specifically `GenericRef`) treats it as an internal ref.  E.g.
```
"$ref" : "./path/to/model.json#/thing"
```
is treated as a relative URI but 
```
"$ref": "path/to/model.json#/thing"
```
is treated as internal.

According to #306, this is intentional behavior in swagger core:
> we added a convenience for a 1.2-style spec, where references did not need the full anchor. 

So the parser needs to account for this by transforming _actual_ relative refs into something that swagger-core will not try to treat as an internal ref.

The fix for #306 added code to make this transformation to relative refs that appear in properties.  The same fix is needed for parameters and schema.

mkistler:
Fixes #421

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
@@ -7,6 +7,7 @@ import io.swagger.models.auth.*;
 import io.swagger.models.parameters.*;
 import io.swagger.models.properties.Property;
 import io.swagger.models.properties.PropertyBuilder;
+import io.swagger.models.properties.RefProperty;
 import io.swagger.util.Json;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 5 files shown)


mkistler:
Okay ... this is harder than I thought.  This code breaks on regular internal refs (starting with a "#").  I will rework and update this PR.

mkistler:
Rework is complete and I've added tests to catch the errors I missed in the earlier commits.  I think this is ready for review now.

mkistler:
@fehguy  What needs to be done to move this forward?

fehguy:
Your fix looks good, I need more time to verify.

mkistler:
Any update on this?

gracekarina:
@mkistler Thanks for this PR! sorry for the delay.

PR #428, opened 2017-03-27
change packaging to (OSGi) bundle

borcsokj:
pull request is created for the change

### DIFF
--- a/modules/swagger-compat-spec-parser/pom.xml
+++ b/modules/swagger-compat-spec-parser/pom.xml
@@ -10,8 +10,22 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>io.swagger</groupId>
     <artifactId>swagger-compat-spec-parser</artifactId>
-    <packaging>jar</packaging>
+    <packaging>bundle</packaging>
     <name>swagger-compat-spec-parser</name>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Export-Package>io.swagger.parser</Export-Package>
+                    </instructions>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
     <dependencies>
         <dependency>
             <groupId>io.swagger</groupId>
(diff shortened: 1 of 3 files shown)


borcsokj:
Create OSGi bundle artifacts instead of plain JAR similar to swagger-core. It is required to use Swagger parser as dependency of Swagger request validator inside Apache Karaf OSGi container installed as feature.

woongiap:
@borcsokj can this be merged already?

borcsokj:
It can be merged but I don't know who can do it/how can I request it.

gracekarina:
@borcsokj @woongiap Thanks for this PR!.
Can you  add a test in OSGi environment?

woongiap:
@gracekarina tested on my local, is that what you meant?

borcsokj:
@gracekarina I also tested changes in OSGi (Apache Karaf) environment. It is not a functional change (that could be covered with unit tests easily) but additional header entries in MANIFEST.MF file of created JAR file.

woongiap:
@gracekarina Hi Grace the change was just making swagger-parser packaging similar to swagger-core and swagger-models, so that it can be used on OSGi environment, can we proceed to merge please?

gracekarina:
Hi, can this be tested with a mock OSCi environment? how do we know for sure that the packaging will behave as expected? Maybe an integration test with embedded OSGi env?

woongiap:
@georgekankava do swagger-core/swagger-mode have similar test?

webron:
@frantuma maybe you have some insight into this one.

frantuma:
@woongiap @borcsokj 

First of all, thank you for this PR. 

Regarding the tests, the idea here is that this is a good moment to add integration tests in OSGi env, as we try to avoid introducing not testable code.

We are aware that this is still the current situation for `swagger-core` but just because we made a mistake before, doesn’t mean we should repeat it.. and providing such tests in the related PR sounds like the best way to do it.

Please let us know if there is a chance anybody out there can indeed add such tests.

cvgaviao:
An osgi manifest is a simple description file that would not require a integration test to it. this PR is taking so long to be merged !

Integration tests on OSGi requires a test framework such as PaxExam. Also it requires a new maven project to hold the integration test code. It is not difficult but has a learning curve and requires time !

Please consider to merge this simple and IMPORTANT change because we (using OSGi) can use swagger-core but unbelievably we can't use swagger-parse easy.  unfortunately, I had to create a bundle to embed the parser v3.

Issue #443, opened 2017-04-25
"minimum" and "maximum" are not imported for definition of type "number" of "integer"


boillodmanuel:
Adding a test case..

PR #444, opened 2017-04-25
fix "minimum" and "maximum" are not imported for definition of type "…

boillodmanuel:
…number" of "integer" (#443)


modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -764,6 +764,7 @@ public void testIssue393() {
         assertNotNull(swagger.getVendorExtensions().get("x-error-defs"));
     }
 
+    @Test
```
boillodmanuel: add forgotten `@Test` annotation

gracekarina:
@boillodmanuel sorry we missed this, can you please check if https://github.com/swagger-api/swagger-parser/pull/619 solved it.

PR #447, opened 2017-04-26
Fix/tests

boillodmanuel:


modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
```diff
@@ -756,6 +756,12 @@ public Model definition(ObjectNode node, String location, ParseResult result) {
                 impl.setUniqueItems(bp);
             }
 
+            BigDecimal bd = getBigDecimal("minimum", node, false, location, result);
```
boillodmanuel: including PR #444 in order to fix the relative path used in #444

After merging the PR #444, I can update this PR to fix the merge conflict (result of the previous merge)

modules/swagger-parser/src/test/java/io/swagger/parser/FileReferenceTests.java
```diff
@@ -11,12 +11,16 @@
...
     @Test
     public void testIssue306() {
-        SwaggerDeserializationResult result = new SwaggerParser().readWithInfo("./src/test/resources/nested-file-references/issue-306.yaml", null, true);
+        SwaggerDeserializationResult result = new SwaggerParser().readWithInfo(getClass().getResource("/nested-file-references/issue-306.yaml").getFile(), null, true);
         assertNotNull(result.getSwagger());
 
         Swagger swagger = result.getSwagger();
 
-        assertTrue(swagger.getDefinitions().size() == 3);
+        assertEquals(swagger.getDefinitions().size(), 5);
```
boillodmanuel: To be reviewed with attention.

The only things that has changed is here. I didn't understand why this test was passing before. As far as I understand, it's normal to find the `Animal` and `Category` definitions in the swagger, but I could be wrong.
I updated the assertions in this way.

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -800,13 +802,65 @@ public void testBadFormat() throws Exception {
         queryParameter = (QueryParameter) parameter;
         assertEquals(queryParameter.getName(), "query-param-collection-format-and-uniqueItems");
         assertEquals(queryParameter.getCollectionFormat(), "multi");
-        assertEquals(queryParameter.isUniqueItems(), true);
+        assertTrue(queryParameter.isUniqueItems());
+    }
+
+    @Test
+    public void testNumberAttributes() throws Exception {
```
boillodmanuel: this test is from PR #444 - I include it in order to fix the relative path used in #444

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerResolverTest.java
```diff
@@ -370,12 +372,11 @@ public void testSettingsAddParametersToEachOperationDisabled() {
                         "        default:\n" +
                         "          description: test response\n";
 
-        SwaggerParser parser = new SwaggerParser();
-        SwaggerDeserializationResult result = parser.readWithInfo(yaml);
 
-        Swagger swagger = result.getSwagger();
+        SwaggerDeserializationResult swaggerDeserializationResult = new SwaggerDeserializer().deserialize(DeserializationUtils.readYamlTree(yaml));
```
boillodmanuel: Actually, using `parser.readWithInfo(yaml)` (done before) already call the SwaggerResolver, so this test was failing. Using `SwaggerDeserializer` directly allow to test `SwaggerResolver` in isolation.

modules/swagger-parser/src/test/resources/number_attributes.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
boillodmanuel: including PR #444 in order to fix the relative path used in #444

gracekarina:
Hi @boillodmanuel  Thanks for this PR, can you please rebase so we can see if this can be merge.
Thanks!

gracekarina:
I believe this tests are fix. Thanks!

Issue #448, opened 2017-04-26
Parsing of Model "example" is not consistent with parsing of Property "example"


boillodmanuel:
Affects v1.0.27

Currently :
- Parsing of Model "example" rely on manual code (see [here](https://github.com/swagger-api/swagger-parser/blob/v1.0.27/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java#L832))
-  Parsing of Model "example" rely jackson deserialization (see [here](https://github.com/swagger-api/swagger-parser/blob/v1.0.27/modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java#L974))

Depending of the type, the parsing of example
- is fine
- is converted to String whereas for property we kept the good type (boolean, array, ...)
- is converted to String with a bug "abc" value become "\"abc\"" with extra double quote (for string)

PR that fix that issue is coming...

boillodmanuel:
PR #449 added

PR #449, opened 2017-04-26
fix Parsing of Model "example" is not consistent with parsing of Prop…

boillodmanuel:
…erty "example" #448


kidnme:
This is a critical fix for me.  What do we have to do to move this along?

DaneWeber:
Does this need to be rebased? Anything else stopping it from being merged?

gracekarina:
@boillodmanuel  @DaneWeber yes this needs to be rebase. Sorry for the delay. will review a soon as rebased 👍

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -468,15 +469,7 @@ public void testIssue292WithCSVCollectionFormat() {
         QueryParameter qp = (QueryParameter) param;
         assertEquals(qp.getCollectionFormat(), "csv");
     }
-
-    @Test
-    public void testIssue255() {
-        SwaggerParser parser = new SwaggerParser();
-
```
gracekarina: Hi, why delete this test instead of making it work with the new approach?.

kidnme: Could I try to make this test work? Again, anything to move this fix along would be great.

gracekarina: yes, please and rebase as well. Thanks!

kidnme: So, grab the pull request, fix the test, rebase against the latest, and resubmit the pull request?  You know I'm not the original author of this pull request ...

gracekarina: oh ok. So how else can we speed this fix?

kidnme: I'll try to patch up the pull request and see if I can make the test work.  Hopefully sometime later today or tomorrow (unless the original author fixes this first).

kidnme: After some work/investigation and consulting the original issue 255, I believe this fix is correct and the original was not (and this is also indicated/hinted at in the fix of 255).  There are unit tests in this fix that verify that a proper ArrayModel is returned for the array example (not a string as in the prior fixes unit test).  I also tested with several downstream pipelines (one that takes swagger -> markdown -> asciidoc -> pdf and a similar one that takes swagger -> markdown -> asciidoc -> static html).  In each case, the fixed version works (i.e. the final pdf / html with an array of objects is displayed correctly).  So, the test was removed because it is somewhat redundant.  Also, it seems like you might want the original issue 255 authors/collaborators to weigh in as I do not know what their downstream usage of the swagger-parser is (so this fix could break their usage - I cannot tell).

kidnme: So, are we good to go?  Will this show up soon in a release?

gracekarina: Hi we need this updated with master branch, I do not have permission to update the @boillodmanuel PR.

gracekarina: @kidnme merged!

kidnme:
So, what causes new releases of the swagger parser?  Just trying to gauge when this fix will make into an official release and into standard repos ...

PR #462, opened 2017-06-02
Add version check to 3.0 parser to prevent 2.0 docs parsed into OpenAPI

arturdzm:
In some conditions 3.0 parser can produce OpenAPI object and it bypasses the 2.0 converter
This lines sometimes produce non null object from v2 documents
`OpenAPI api = mapper.readValue(swaggerAsString, OpenAPI.class);
result.setOpenAPI(api);`

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
@@ -1,5 +1,8 @@
 package io.swagger.parser.v3;
-
+import com.fasterxml.jackson.core.JsonFactory;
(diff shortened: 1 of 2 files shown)


fehguy:
Thank you--can you please add a test to catch the failing condition?

arturdzm:
Added test to check that v3 parser returns null if v2 document is supplied

fehguy:
I've looked into this--the correct way to handle the version check during parsing is inside the v3 parser after turning into a JsonNode.  Please see upcoming PR.

arturdzm:
@fehguy Hey, I though above using JsonNode at first, but did streaming parsing approach to optimize parsing to skip over children if openapi key is found. JsonNode would parse entire document each time. I think it might be a bit slower with large documents. And in most documents this key is at the top

PR #466, opened 2017-06-07
Update json-schema-validator (SONATYPE-2015-0090)

msymons:
Update json-schema-validator from 2.2.3 to 2.2.8 (with change in groupId to reflect ownership change).   Addresses Cross Site Scripting vulnerability from libphonenumber dependency.


fehguy:
@webron is the license still OK for this dependency?

webron:
Yeah, the license is the same as before with the version. Dual license - LGPLv3, APLv2 - obviously we'd assume APLv2 for our needs.

fehguy:
thank @webron

PR #470, opened 2017-06-12
primitive schemas are no longer lost in conversion

bedag-moo:
Fixes https://github.com/swagger-api/swagger-codegen/issues/1241

### DIFF
--- a/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
+++ b/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
@@ -331,7 +331,9 @@ public class SwaggerCompatConverter implements SwaggerParserExtension {
                 output = am;
             } else {
                 Property input = PropertyBuilder.build(type, format, null);
-                if (input == null && !"void".equals(type)) {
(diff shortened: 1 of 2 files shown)


fehguy:
Hi @bedag-moo can you please supply a test for this?

bedag-moo:
The best test would probably be an integration test with swagger-codegen, but I don't see any infrastructure for automating such tests in swagger-parser? What kind of test are you looking for?

fehguy:
Typically when using codegen, there is a json or yaml specification file, which is run through the parser, that has some situation that the parser doesn't like.  In the case of this PR, I'd like to see a small specification with a broken behavior which is fixed by this PR.  Then we know precisely what was causing the issue, and are protected against it happening again.  So my guess is that your specification file will be enough for us to figure out what to test.  Can you share it here or in private?

bedag-moo:
@fehguy, do you need anything else from me so you can merge this?

fehguy:
Well, your spec may exhibit the issue, but until there's an automated unit test we can't merge this.  So it can be done faster by you, or you can wait for us to do it.

bedag-moo:
Here you go.

fehguy:
Thank you

PR #478, opened 2017-07-03
Adding a call to the resolver in the parser

gracekarina:
Added a call to the OpenAPIResolver on the OpenAPIParserV3 class, and added a test or it.


modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -27,12 +27,11 @@
     public SwaggerParseResult readLocation(String url, List<AuthorizationValue> auth, ParseOptions options) {
         SwaggerParseResult result = new SwaggerParseResult();
         try {
-            // TODO
             JsonNode node = YAML_MAPPER.readValue(new URL(url), JsonNode.class);
             if(node != null && node.get("openapi") != null) {
                 JsonNode version = node.get("openapi");
                 if(version.asText() != null && version.asText().startsWith("3.0")) {
-                    result = new OpenAPIDeserializer().deserialize(node);
+                    result.setOpenAPI(new OpenAPIResolver(new OpenAPIDeserializer().deserialize(node).getOpenAPI(),null,null).resolve());
```
fehguy: You need to pass the `auth` option along when doing the resolving, also consider the `options` in the call

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -55,8 +54,12 @@ public SwaggerParseResult readContents(String swaggerAsString, List<Authorizatio
                 mapper = YAML_MAPPER;
             }
             try {
-                OpenAPI api = mapper.readValue(swaggerAsString, OpenAPI.class);
-                result.setOpenAPI(api);
+                OpenAPIDeserializer deserializer = new OpenAPIDeserializer();
+                JsonNode rootNode = mapper.readTree(swaggerAsString.getBytes());
+                SwaggerParseResult  deserializeOpenAPI = deserializer.deserialize(rootNode);
+                OpenAPIResolver resolver = new OpenAPIResolver(deserializeOpenAPI.getOpenAPI(),null,null);
```
fehguy: Same with previous comment, handle `auth` and `options`

modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OAIDeserializationTest.java
```diff
@@ -10,26 +11,25 @@
     @Test
     public void testDeserializeSimpleDefinition() throws Exception {
         String yaml =
```
fehguy: seems like this variable should be called `json` because it 'aint yaml...

modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OpenAPIParserV3Test.java
```diff
@@ -0,0 +1,141 @@
+package io.swagger.parser.test;
+
```
fehguy: Be sure to add some tests that pass `auth` and `ParseOptions` because those are needed for proper resolving

gracekarina: I added the test.

modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OAIDeserializationTest.java
gracekarina: ^_^ . Changed it.

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
gracekarina: Ok. I added the options and the authorization List

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
gracekarina: Ok. I added the options and the authorization List

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -27,12 +27,18 @@
...
+                    auth = new ArrayList<>();
+                }
                 if(version.asText() != null && version.asText().startsWith("3.0")) {
-                    result = new OpenAPIDeserializer().deserialize(node);
+                    if (options != null) {
+                        if (options.isResolve()) {
+                            result.setOpenAPI(new OpenAPIResolver(new OpenAPIDeserializer().deserialize(node).getOpenAPI(), auth, null).resolve());
+                        }
+                    }
                 }
```
fehguy: I think you're missing the `else` here.  If you don't resolve, you can still deserialize and set it in the object.

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -54,12 +60,22 @@ public SwaggerParseResult readContents(String swaggerAsString, List<Authorizatio
...
+                        OpenAPIDeserializer deserializer = new OpenAPIDeserializer();
+                        JsonNode rootNode = mapper.readTree(swaggerAsString.getBytes());
+                        SwaggerParseResult deserializeOpenAPI = deserializer.deserialize(rootNode);
+                        OpenAPIResolver resolver = new OpenAPIResolver(deserializeOpenAPI.getOpenAPI(), auth, null);
+                        result.setOpenAPI(resolver.resolve());
+
+                    } catch (Exception e) {
+                        result.setMessages(Arrays.asList(e.getMessage()));
+                    }
+                }
```
fehguy: Same with above, please use the deserializer if resolve is not set, or if options are null

gracekarina: Ok. Changes made

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
gracekarina: Ok. Changes made

PR #493, opened 2017-07-25
Maven plugin for spec reading and validation

BartlomiejLewandowski:
So I've started to integrate swagger generation into our build process. I've noticed that there is no package or method which allows you to verify if a swagger file is valid - that each operation has a response and that there are no definition references which cannot be resolved.

So the PR adds a new module which creates a maven plugin with a new mojo called validate. It throws a build error exception when swagger parser returns at least one error.

Do you have any advice on where to place the validation code? I know this can be improved, but wanted to quickly code this and get your feedback.


BartlomiejLewandowski:
I've decided to not pursue this approach, found a node module which does schema and spec validation, so I've used a maven frontend plugin to take care of this.

Issue #499, opened 2017-08-17
Parsing of boolean `additonalProperties` inconsistent with OpenAPI 3.0.0 specification


mbeko:
I'm used to setting `"additionalProperties": false` from JsonSchema, hence did so in an OpenAPI 3.0.0 file, especially as it's [allowed by the specification](https://swagger.io/specification/#properties-99).

Using version `2.0.0-SNAPSHOT` of `swagger-parser-v3` as found on oss.sonatype.org for OpenAPI file validation, I get a parse result message saying that `additonalProperties` is not of type `object`.

I'm aware of issue #366, but I think either
* the parser should accept boolean values for `additionalProperties` as well
or
* it must be clearly stated in the specification if the parser is supposed to frown upon such values.

In addition, the message is a bit misleading because it is reported only once and only for the top level of the object definiton, even though there were several occurences of `"additonalProperties": false` on deeper levels of this definition. Providing a more precise location would have helped a lot with debugging this issue.

webron:
Thanks for the feedback, @mbeko. The parser should definitely allow it.
Can you provide us with a sample spec to help test it?

Better yet, if you can submit a PR (against the 2.0 branch) adding a test to the test suite, we'll make sure it passes.

mbeko:
Great, thanks! I'll reduce my specification to a minimal example exposing the problem and submit a pull request.

Since I should do it against the `2.0` branch: Should I keep the specification in the 3.0.0 format or rewrite it as 2.0?

webron:
Should be 3.0.0. 

2.0 did not allow boolean values to `additionalProperties`.

slinkydeveloper:
What about that bug?

mbeko:
I have added the sample specification and the test, but haven't received feedback since then.

Concerning news about the overall adaptation of the tools to 3.0.0, [issue #4669 on swagger-codegen](https://github.com/swagger-api/swagger-codegen/issues/4669) indicates that it lags behind the specification.

webron:
Sorry about this - just got lost in the mass emails. We'll try to take a look shortly.

mbeko:
No problem and thank you :)

slinkydeveloper:
There's a chance to get this fix on upstream before end January? :smile:

gracekarina:
Hi, please checkout the last RC, this was covered there.

twof:
Hi! This bug appears to have returned in a recent release of swagger-parser. I'll open a new issue.

PR #500, opened 2017-08-18
Add test for boolean `additionalProperties` (#499)

mbeko:
Provides a sample file and test which reproduces issue #499 stating that according to the OpenAPI 3.0.0 specification, `additionalProperties` of boolean type should be allowed.

Currently, only removing the boolean `additionalProperties` makes this test pass. You will also notice that the error message only reports an error at the location `components.schemas'.someObject'` even if only more deeply nested boolean `additionalProperties` are left.


gracekarina:
@mbeko thanks for the test, the solution for this is merged, please checkout `2.0`branch

PR #501, opened 2017-08-18
adding shell methods and test

gracekarina:
Added method read, to used it in the readme file.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
@@ -47,11 +47,12 @@ public class OpenAPIV3Parser implements SwaggerParserExtension {
                 }
                 if (version != null && version.startsWith("3.0")) {
                     if (options != null) {
+                        OpenAPIResolver resolver = new OpenAPIResolver(result.getOpenAPI(), auth, null);
                         if (options.isResolve()) {
-                            result.setOpenAPI(new OpenAPIResolver(result.getOpenAPI(), auth, null).resolve());
+                            result.setOpenAPI(resolver.resolve());
                         }
                         if (options.isResolveFully()) {
-                            result.setOpenAPI(new OpenAPIResolver(result.getOpenAPI(), auth, null).resolve());
+                            result.setOpenAPI(resolver.resolve());
                             new ResolverFully().resolveFully(result.getOpenAPI());
                         }
                     }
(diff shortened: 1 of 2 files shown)


modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OpenAPIV3ParserTest.java
```diff
@@ -183,6 +184,12 @@ public void test30NoOptions(@Injectable final List<AuthorizationValue> auths) th
         Assert.assertEquals(result.getOpenAPI().getComponents().getSchemas().get("OrderRef").getType(),"object");
     }
 
+    @Test
+    public void testShellMethod(){
+        OpenAPI openAPI = new OpenAPIV3Parser().read("https://gist.githubusercontent.com/webron/e3b0650dfcc06fe8236841fe599c287f/raw/12512eb5343dd56ce79369d7ff58072584bd0dc7/openapi.yaml");
```
HugoMario: It's not so good to have remote calls on tests, since the idea is to execute them in local enviroments. I suggest to mock this call. So you can evaluate the behavior you expect without making remote calls.

gracekarina: Oki

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -39,17 +40,20 @@ public SwaggerParseResult readLocation(String url, List<AuthorizationValue> auth
...
-                        new ResolverFully().resolveFully(result.getOpenAPI());
+            if (result.getOpenAPI() != null) {
+                String version = result.getOpenAPI().getOpenapi();
+                if (auth == null) {
+                    auth = new ArrayList<>();
+                }
+                if (version != null && version.startsWith("3.0")) {
+                    if (options != null) {
+                        if (options.isResolve()) {
+                            result.setOpenAPI(new OpenAPIResolver(result.getOpenAPI(), auth, null).resolve());
```
HugoMario: can you update this block tu avoid double call on OpenAPIResolver?

HugoMario:
I added some comments related to add soem tweaks to code, it would be good to have a logger to write possible exceptions.

modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -39,17 +40,20 @@ public SwaggerParseResult readLocation(String url, List<AuthorizationValue> auth
...
-                auth = new ArrayList<>();
-            }
-            if(version != null && version.startsWith("3.0")) {
-                if (options != null) {
-                    if (options.isResolve()) {
-                        result.setOpenAPI(new OpenAPIResolver(result.getOpenAPI(), auth, null).resolve());
-                    }if(options.isResolveFully()){
-                        result.setOpenAPI(new OpenAPIResolver(result.getOpenAPI(), auth, null).resolve());
-                        new ResolverFully().resolveFully(result.getOpenAPI());
+            if (result.getOpenAPI() != null) {
```
HugoMario: The statement `result.getOpenAPI()` is called many times, it would be better to make only one call, assigning to a variable and use this variable instead.

PR #511, opened 2017-08-29
Fix example implementation on schema, header, parameter and media Type

gracekarina:
The Example attribute now supports primitives (integer, big decimal, string), objects and arrays, where the spec says the value can be Any. 
Please Merge [[https://github.com/swagger-api/swagger-core/pull/2406]](https://github.com/swagger-api/swagger-core/pull/2406) before merging this.

### DIFF
--- a/modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OpenAPIV3ParserTest.java
+++ b/modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OpenAPIV3ParserTest.java
@@ -155,7 +155,7 @@ public class OpenAPIV3ParserTest {
 
         Assert.assertNotNull(result);
         Assert.assertNotNull(result.getOpenAPI());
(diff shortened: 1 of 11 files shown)


modules/swagger-parser-v3/src/test/java/io/swagger/parser/test/OpenAPIV3ParserTest.java
```diff
@@ -155,15 +155,14 @@ public void test30(@Injectable final List<AuthorizationValue> auths) throws Exce
 
         Assert.assertNotNull(result);
         Assert.assertNotNull(result.getOpenAPI());
-        Assert.assertEquals(result.getOpenAPI().getOpenapi(), "3.0.0-RC1");
+        Assert.assertEquals(result.getOpenAPI().getOpenapi(), "3.0");
```
webron: This should be `3.0.0`?

PR #520, opened 2017-09-10
code refactoring to fix #519

### LINKED ISSUE #519: ReadContents and ReadLocation have not the same behavior with OAS 3.0.0
Hello,

parsing a OAS 3.0.0 spec with this configuration :
```yaml
openapi: '3.0.0'
components:
  examples:
    ANull:
      value: null
```

Then
if I use `OpenAPIV3Parser.readContents`, the `ANull` JsonNode looks like `{"value":null}`.
if I use `OpenAPIV3Parser.readLocation`, the `ANull` JsonNode looks like `{}`.

The "bug" is (obviously 😄) in `io.swagger.parser.v3.OpenAPIV3Parser` class.
`OpenAPIV3Parser.readContents` uses

```java
static {
        JSON_MAPPER = ObjectMapperFactory.createJson();
        YAML_MAPPER = ObjectMapperFactory.createYaml();
}
```

wheras `OpenAPIV3Parser.readLocation` uses

```java
ObjectMapper mapper = Json.mapper(); //for JSON
DeserializationUtils.readYamlTree(data); //for YAML
```

And it seems that they don't have the same behavior.

I'm preparing a PR to propose a refactoring of this class.

phiz71:


modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/OpenAPIV3Parser.java
```diff
@@ -37,161 +37,96 @@
...
         }
-
         catch (Exception e) {
-            result.setMessages(Arrays.asList(e.getMessage()));
+            result.setMessages(Arrays.asList("unable to read location `" + url + "`"));
         }
         return result;
     }
 
-    public OpenAPI read(String location) {
```
gracekarina: Hi, Why remove this method?

phiz71: Hi,
IMHO this method is not useful. `readContents` and `readLocation` are enough. But if `read(location)` has to be maintened, maybe should we call it `readAndResolve` or `readWithResolve` ?

And about the other `read` method, it is a duplicate of the `io.swagger.parser.OpenAPIParser`. This class is already a `SwaggerExtension`, to my mind, it doesn't have to know the others existing extensions (And maybe I should have remove `getExtensions` and `transform` too in my PR)

Finally, we could just have :
```java
public SwaggerParseResult readLocation(String url, List<AuthorizationValue> auth, ParseOptions options) {...}
public SwaggerParseResult readContents(String swaggerAsString, List<AuthorizationValue> auth, ParseOptions options) {...}

private String getDataFromLocation(String location, List<AuthorizationValue> auths) throws Exception {...}
private JsonNode mapDataToJsonNode(String data) throws JsonProcessingException, IOException {...}
private SwaggerParseResult readWithInfo(JsonNode node){...}

//and for instance
public OpenAPI readAndResolve(String location) {
    options.setResolve(true);
    return readLocation(location, null, options).getOpenAPI();
}
```

But this is just my opinion :smile:

phiz71:
I had a look at the others classes of the others modules. I think I will cancel this PR and propose something more simple.
Finally, the only thing that is really important for me, is to use the same JSON or YAML parser whether you use readContents or readLocation.
I will work on it next days.

phiz71:
As I said, a simplier PR :)

webron:
Thanks, @phiz71!

PR #547, opened 2017-10-12
Property description is lost when property uses ref schema

gdombiak:
Noticed that properties with inline schema keep their description just fine but it was not working for properties with a ref schema.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/parser/v3/util/OpenAPIDeserializer.java
@@ -1823,6 +1823,12 @@ public class OpenAPIDeserializer {
             schema = SchemaTypeUtil.createSchemaByType(node);
         }
 
+        // Set description for inline or ref schemas
+        String value = getString("description",node,false,location,result);
+        if (StringUtils.isNotBlank(value)) {
+            schema.setDescription(value);
+        }
+
         JsonNode ref = node.get("$ref");
         if (ref != null) {
             if (ref.getNodeType().equals(JsonNodeType.STRING)) {
@@ -1835,7 +1841,7 @@ public class OpenAPIDeserializer {
         }
 
 
-        String value = getString("title",node,false,location,result);
+        value = getString("title",node,false,location,result);
         if (StringUtils.isNotBlank(value)) {
             schema.setTitle(value);
         }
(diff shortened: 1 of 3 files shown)


webron:
@gdombiak technically, any property alongside a $ref should be lost by definition, so not sure this fix makes sense to merge.

gdombiak:
Hi Ron,

I was not able to find in the spec that any other attribute next to $ref should be ignored. Where is this in the spec?

Here is an example that illustrates our existing need. 

```
openapi: 3.0.0
info:
  title: Pets API
  version: '1.0'
paths:
  /pets/{id}:
    get:
      responses:
         '200':
          description: Ok
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"
components:
  schemas:
    Pet:
      properties:
        currentAddress:
          description: Address where the pet currently lives
          $ref: '#components/schemas/Address'
        previousAddress:
          description: Address where the pet used to live
          $ref: '#components/schemas/Address'
    Address:
      properties:
        address1:
          type: string
```

As you can see, currentAddress and previousAddress are both referencing the same schema but each one needs its own description to explain what the property is about. What would be another way of solving this according to the spec?

Thanks,
Gaston

webron:
Take a look at https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#referenceObject - notice what it says under the `Fixed Fields` table.

The only solution is an ugly one - using `allOf`, one with the description and one with the $ref.

gdombiak:
Hi Ron,

Thanks so much for pointing to the relevant part of the spec that covers this. That helps a lot.

I agree that using allOf is not a clear and simple solution and possibly confusing. Here is an example of how both options would look like:

```
        previousAddress:
          allOf:
            - type: object
              description: Address where the pet used to live
            - $ref: '#components/schemas/Address'
```

vs

```
        previousAddress:
            description: Address where the pet used to live
            $ref: '#components/schemas/Address'
```

Having to use an allOf with a real ref schema and another object schema with no properties but a description is confusing. Do you think this is something we can improve in the spec? Maybe something like this?
```
This object cannot be extended with additional properties and any properties added SHALL be ignored with the exception of description.
```

Happy to work with you on a proposal to improve the spec if you think that is possible. I think that standards that make it simple for people to use help increase adoption. :)

Toda,
Gaston

webron:
You don't need to convince me when it comes to the limitations of JSON Reference. However, we do follow the JSON Reference spec on which we have no influence. There's already long, long (and I mean, really long) discussion on it over at https://github.com/OAI/OpenAPI-Specification/issues/556.

gdombiak:
Closing (abandoning) PR. Ron's comment are spot on. We can revisit this after JSON Schema allows description in ref elements.

PR #555, opened 2017-10-18
Issue-35 Fixed

ralphdoe:
Fixed: https://github.com/swagger-api/swagger-spec-converter/issues/35

### DIFF
--- a/modules/swagger-parser-v2-converter/src/test/resources/issue-33.json
+++ b/modules/swagger-parser-v2-converter/src/test/resources/issue-33.json
@@ -1,0 +1,68 @@
+{
+    "swagger": "2.0",
+    "info": {
+        "version": "1.0",
+        "title": "Examples"
+    },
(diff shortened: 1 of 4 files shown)


webron:
Does this affect only parameters? It should affect schemas/models in general.

modules/swagger-parser-v2-converter/src/test/resources/issue-33.json
```diff
@@ -0,0 +1,68 @@
...
+                                "foo": "bar",
+                                "random": "example"
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    },
+    "definitions": {
```
webron: Can you add a case to the test where one of the definitions has a property with `x-nullable`?

ralphdoe: Checked! Test case added!

PR #568, opened 2017-11-03
Added flag to avoid aggregation of anyOf/allOf/oneOf schemas while using resolveFully()

slinkydeveloper:
Some users don't want to aggregate anyOf/allOf/oneOf schemas but simply wants all refs solved.
This solution adds a flag inside the `ResolverFully` and when schemas are processed It checks this flag. In case user don't want to aggregate, It generates a new list of subschemas solved and saves it to the composedSchema object.

I've also added a flag in `ParseOptions` that `OpenAPIV3Parser` reads during `readContents()` and I've wrote a test. For default the parser aggregate, as is now

### DIFF
--- a/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIResolverTest.java
+++ b/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIResolverTest.java
@@ -540,9 +540,21 @@ public class OpenAPIResolverTest {
         options.setResolveFully(true);
         options.setResolveCombinators(false);
 
+        // Testing components/schemas
         OpenAPI openAPI = new OpenAPIV3Parser().readLocation("src/test/resources/composed.yaml",auths,options).getOpenAPI();
 
         ComposedSchema allOf = (ComposedSchema) openAPI.getComponents().getSchemas().get("ExtendedAddress");
+        assertEquals(allOf.getAllOf().size(), 2);
+
+        assertTrue(allOf.getAllOf().get(0).getProperties().containsKey("street"));
+        assertTrue(allOf.getAllOf().get(1).getProperties().containsKey("gps"));
+
+        // Testing path item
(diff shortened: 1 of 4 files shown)


slinkydeveloper:
@webron can you accept this pr? I need that for some bugs inside vertx-web (https://github.com/vert-x3/vertx-web/issues/733 and https://github.com/vert-x3/vertx-web/issues/728)

vinscom:
There are too many diff. just because of space. Can you please just have your changes visible as part of diff.

gracekarina:
Hi! @slinkydeveloper  what do you mean, when users don't want to aggregate them? It looks to me that when I use aggregateCombinators: false, the ref is resolved, and when it's true it isn't (Shouldn't it be the other way around?). So I'm sorry if I don't get what is the use case of this. Can you please explain to me when will this be use?. Thanks!

slinkydeveloper:
I've modified a bit the test. Now it checks also the path @gracekarina suggested :smile:

modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIResolverTest.java
```diff
@@ -534,6 +534,34 @@ public void selfReferenceTest(@Injectable final List<AuthorizationValue> auths)
...
+
+        // Testing path item
+        ComposedSchema schema = (ComposedSchema) openAPI.getPaths().get("/withInvalidComposedModel").getPost().getRequestBody().getContent().get("application/json").getSchema();
+
+        // In fact the schema resolved previously is the same of /withInvalidComposedModel
+        assertEquals(schema, allOf);
+
+        assertEquals(allOf.getAllOf().size(), 2);
+
+        assertTrue(allOf.getAllOf().get(0).getProperties().containsKey("street"));
```
gracekarina: this should be schema.getAllOf...

slinkydeveloper: Ops, brutal copy paste :flushed:

gracekarina:
Thanks @slinkydeveloper

PR #578, opened 2017-11-13
Referring without components

joeljons:
According to how i interpret the spec the "components" object is not required. I wish it were! (Or at least default to an empty components object) I have had so some bugs where I assume components exists...

According to the deserializer it assumes it is required. (The `true` means it is required) https://github.com/swagger-api/swagger-parser/blob/647bfe2c8cad362d837c052980939134933c70b2/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java#L137

This PR fixes a NPE in the resolver when the spec is missing the components tag and using external refs. I haven't checked, but I wouldn't be surprised if there are other cases of `openAPI.getComponents().getSomeSubObject()` assuming components is not null. Maybe thinking about defaulting the root components to an empty object is worth thinking about...

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
@@ -154,37 +154,37 @@ public class ResolverCache {
     private Object loadInternalRef(String ref) {
         Object result = null;
 
-        if(ref.startsWith("#/components/schemas")) {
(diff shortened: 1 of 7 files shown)


joeljons:
@gracekarina Do you want me to do that here in this PR? Or will you do it separately?

joeljons:
@gracekarina It wasn't that many places so I went ahead and did it.

gracekarina:
Thanks @joeljons

PR #580, opened 2017-11-13
Fix to SwaggerCompatConverter so it can load inputSpec from classpath.

Dema:
Swagger20Parser already can load input spec from classpath, but codegen project loads extensions one them is SwaggerCompatConverter and it fails loading from classpath

This fixes https://github.com/swagger-api/swagger-codegen/issues/2666

### DIFF
--- a/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
+++ b/modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
@@ -52,18 +52,23 @@ import io.swagger.models.resourcelisting.BasicAuthorization;
 import io.swagger.models.resourcelisting.ImplicitGrant;
 import io.swagger.models.resourcelisting.OAuth2Authorization;
 import io.swagger.models.resourcelisting.ResourceListing;
+import io.swagger.parser.util.ClasspathHelper;
 import io.swagger.parser.util.RemoteUrl;
 import io.swagger.parser.util.SwaggerDeserializationResult;
 import io.swagger.report.MessageBuilder;
 import io.swagger.transform.migrate.ApiDeclarationMigrator;
 import io.swagger.transform.migrate.ResourceListingMigrator;
 import io.swagger.util.Json;
+import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


webron:
Thanks for the PR. Can you add a test to it?

modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
```diff
@@ -186,7 +191,22 @@ public ResourceListing readResourceListing(String input, MessageBuilder messages
                 String json = RemoteUrl.urlToString(input, auths);
                 jsonNode = Json.mapper().readTree(json);
             } else {
-                jsonNode = Json.mapper().readTree(new File(input));
+                final String fileScheme = "file:";
```
vinscom: Its final string. Should be constant.

modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
```diff
@@ -186,7 +191,22 @@ public ResourceListing readResourceListing(String input, MessageBuilder messages
...
+                java.nio.file.Path path;
+                if (input.toLowerCase().startsWith(fileScheme)) {
```
vinscom: To avoid NPE, to compare with fileSchema, as it is constant

modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
```diff
@@ -186,7 +191,22 @@ public ResourceListing readResourceListing(String input, MessageBuilder messages
...
+                    path = Paths.get(URI.create(input));
+                } else {
+                    path = Paths.get(input);
+                }
+                String json;
+                if (Files.exists(path)) {
+                    json= FileUtils.readFileToString(path.toFile(), "UTF-8");
```
vinscom: StandardCharsets already has "UTF-8" constant

modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
```diff
@@ -489,7 +509,21 @@ public ApiDeclaration readDeclaration(String input, MessageBuilder messages, Lis
                 String json = RemoteUrl.urlToString(input, auths);
                 jsonNode = Json.mapper().readTree(json);
             } else {
-                jsonNode = Json.mapper().readTree(new java.io.File(input));
+                final String fileScheme = "file:";
```
vinscom: Any reason for duplicate code ? Can't we refactor this code into function

modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
Dema: No, it can't throw NPE on line 196, it'll throw it in the previous if(input.startsWith("http")) statement.

modules/swagger-compat-spec-parser/src/main/java/io/swagger/parser/SwaggerCompatConverter.java
Dema: I took this exact code from https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser/src/main/java/io/swagger/parser/Swagger20Parser.java#L46
Maybe it makes sense to refactor this code too?

gracekarina:
@Dema thanks, sorry for the delay on this.

PR #590, opened 2017-11-22
$ref name conflicts for Swagger 2

joeljons:
Create new names by adding _{number} to new ref name if a conflict would occur.


gracekarina:
@joeljons Hi, is there a specific case for this to be done?

joeljons:
In the case of a $ref:d spec contains the same definition name (with different content) as an already existing definition. Also if multipe $ref:d contains the same definition names (with different content).

PR #591, opened 2017-11-22
$ref name conflicts for OAS 3

joeljons:
Create new names by adding _{number} to new ref name if a conflict would occur.


gracekarina:
Hi, @joeljons is this a bug that's being fix, or a better way of handling the same names in the schemas? :)

joeljons:
This is a bug fix for the case when an already existing schema has the same name as an externally referred schema. This collision wasn't handled.

Issue #342, opened 2016-11-05
$ref in additionalProperties in external yaml ignored


rakus:
Initially opened against swagger codegen as [#4135](https://github.com/swagger-api/swagger-codegen/issues/4135). They send me here.

It seems that swagger-parser cannot handle a $ref below `additionalProperties` in a external yaml file.

Example:

---
swagger.yaml
```yaml
swagger: "2.0"
info:
  title: Test API
  version: 1.0.0
host: swagger.io
basePath: /show-bug
schemes:
- http
produces:
- application/json
consumes:
- application/json

paths:
  /tests:
    get:
      operationId: listTests
      responses:
        200:
          description: Returns results.
          schema:
              $ref: "./globals.yaml#/definitions/result"
```

---
globals.yaml
```yaml
definitions:
  link-object:
    type: object
    additionalProperties:
      $ref: '#/definitions/rel-data'

  rel-data:
    type: object
    required:
      - href
    properties:
      href:
        type: string
      note:
        type: string

  result:
    type: object
    properties: 
      name:
        type: string
      _links:
        $ref: "#/definitions/link-object"
```

After parsing the object `rel-data` (`globals.yaml#/definitions/rel-data`) is missing.
The verbose output of swagger-codegen shows it:

```JSON
{
  "swagger" : "2.0",
  "info" : {
    "version" : "1.0.0",
    "title" : "Test API"
  },
  "host" : "swagger.io",
  "basePath" : "/show-bug",
  "schemes" : [ "http" ],
  "consumes" : [ "application/json" ],
  "produces" : [ "application/json" ],
  "paths" : {
    "/tests" : {
      "get" : {
        "operationId" : "listTests",
        "parameters" : [ ],
        "responses" : {
          "200" : {
            "description" : "Returns results.",
            "schema" : {
              "$ref" : "#/definitions/result"
            }
          }
        }
      }
    }
  },
  "definitions" : {
    "result" : {
      "type" : "object",
      "properties" : {
        "name" : {
          "type" : "string"
        },
        "_links" : {
          "$ref" : "#/definitions/link-object"
        }
      }
    },
    "link-object" : {
      "type" : "object",
      "additionalProperties" : {
        "$ref" : "#/definitions/rel-data"
      }
    }
  }
}
```

fehguy:
This is not implemented

jimschubert:
[...]

PR #605, opened 2017-12-16
Schema with Additional properties was not solving remote refs #342

gracekarina:
fixing bug when parser trying to solve remote ref in additional properties schema issue #342

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -111,6 +111,26 @@ public final class ExternalRefProcessor {
                     }
                 }
             }
+            if(schema.getAdditionalProperties() != null){
+                Schema additionalProperty = schema.getAdditionalProperties();
+                if (additionalProperty.get$ref() != null) {
+                    processRefProperty(additionalProperty, file);
+                } else if (additionalProperty instanceof ArraySchema) {
+                    ArraySchema arrayProp = (ArraySchema) additionalProperty;
+                    if (arrayProp.getItems().get$ref() != null) {
+                        processRefProperty(arrayProp.getItems(), file);
+                    }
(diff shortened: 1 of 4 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
```diff
@@ -111,6 +111,26 @@ public String processRefToExternalSchema(String $ref, RefFormat refFormat) {
                     }
                 }
             }
+            if(schema.getAdditionalProperties() != null){
+                Schema additionalProperty = schema.getAdditionalProperties();
+                if (additionalProperty.get$ref() != null) {
+                    processRefProperty(additionalProperty, file);
+                } else if (additionalProperty instanceof ArraySchema) {
+                    ArraySchema arrayProp = (ArraySchema) additionalProperty;
+                    if (arrayProp.getItems().get$ref() != null) {
```
HugoMario: how are we sure that "getItems()" won't return null?

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
```diff
@@ -111,6 +111,26 @@ public String processRefToExternalSchema(String $ref, RefFormat refFormat) {
...
+                        processRefProperty(arrayProp.getItems(), file);
+                    }
+                } else if (additionalProperty.getAdditionalProperties() != null) {
+                    Schema mapProp =  additionalProperty;
+                    if (mapProp.getAdditionalProperties().get$ref() != null) {
+                        processRefProperty(mapProp.getAdditionalProperties(), file);
+                    } else if (mapProp.getAdditionalProperties() instanceof ArraySchema &&
+                            ((ArraySchema) mapProp.getAdditionalProperties()).getItems().get$ref() != null) {
```
HugoMario: same question here.

PR #612, opened 2017-12-28
Supporting change in additionalProperties Boolean and Schema

gracekarina:
refs swagger-api/swagger-core#2461
#499 and fixing refs in additionalProperties Resolving RefExternalResolver #342 (in v3)

### DIFF
--- a/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
+++ b/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
@@ -22,16 +22,15 @@ import io.swagger.models.parameters.RefParameter;
 import io.swagger.models.parameters.SerializableParameter;
 import io.swagger.models.properties.AbstractNumericProperty;
 import io.swagger.models.properties.ArrayProperty;
-import io.swagger.models.properties.IntegerProperty;
+import io.swagger.models.properties.FileProperty;
 import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.ObjectProperty;
 import io.swagger.models.properties.Property;
 import io.swagger.models.properties.RefProperty;
-import io.swagger.models.properties.StringProperty;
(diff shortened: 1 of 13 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/InlineModelResolver.java
```diff
@@ -454,11 +454,15 @@ public Schema modelFromProperty(Schema object, @SuppressWarnings("unused") Strin
...
             example = obj.toString();
         }
-
+        Schema items;
         ArraySchema model = new ArraySchema();
         model.setDescription(description);
         model.setExample(example);
-        model.setItems(object.getAdditionalProperties());
+        if (object.getAdditionalProperties() != null && object.getAdditionalProperties() instanceof Schema) {
+            items = (Schema)  object.getAdditionalProperties();
```
ralphdoe: It seems the variable item was created only for assignation, if there's no validation or another it might be a good idea to delete it and make the setItems directly model.setItems((Schema)object.getAdditionalProperties());

gracekarina: fixed!

modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
```diff
@@ -1007,50 +1022,58 @@ public Schema convert(io.swagger.models.Model v2Model) {
...
+                return result;
+
+        } else if (v2Model instanceof RefModel){
+            RefModel model = (RefModel) v2Model;
+            result =  new Schema().$ref(model.get$ref());
+            return result;
         }
 
-        return result;
+        return null;
```
ralphdoe: You can return result instead null.

gracekarina:
This PR depends on https://github.com/swagger-api/swagger-core/pull/2598

PR #613, opened 2017-12-31
Fix discriminator mapping If there are any name conflict for ComposedSchema (oneOf)

adarshsharma:
in oneOf ComposedSchema by default the propertyName field of the discriminator has value same as schema name. In case of name conflict the name of the schema changes, effectively changing the mapping. Fixed this case by keeping the older mapping.
Also in case of name conflict, the reference in the mapping was not changing.


gracekarina:
@adarshsharma  Hi, sorry for the delay, can you please update this PR so I can merge.
Thanks!

adarshsharma:
Sorry I deleted the branch by mistake. The updated changes are in PR: https://github.com/swagger-api/swagger-parser/pull/631

PR #614, opened 2018-01-02
Avoid NPE when parameters only exist on path level

joeljons:
If a method had no parameters we got NPE when resolving


gracekarina:
Thanks! @joeljons

Issue #616, opened 2018-01-07
allOf fails to resolve RELATIVE ref


EtonDoze:
Here is an example
```
.
├── models
│   ├── pet.json
│   └── fancy_pet.json    #inherits pet.json using allOf
└── parent.json
```
### parent.json
```
{
  "swagger": "2.0",
  "paths": {
    "/pet": {
      "get": {
        "responses": {
          "200": {
            "schema": {
              "$ref": "./models/pet.json"
            }
          }
        }
      }
    },
    "/fancy-pet": {
        "responses": {
          "200": {
            "schema": {
              "$ref": "./models/fancy_pet.json"
            }
          }
        }
      }
    }
  }
}
```

### models/`pet.json`
```
{
  "type": "object",
  "properties": {
    "id": {
      "type": "integer",
      "format": "int64"
    },
    "name": {
      "type": "string",
      "example": "Cooper"
    }
  }
}
```

### models/`fancy_pet.json`
Notice here that `pet.json` is referenced using relative path which is in the same `models` folder.
```
{
  "allOf": [
    {
      "$ref": "./pet.json"
    },
    {
      "properties": {
        "breed": {
          "type": "string",
          "example": "Labrador"
        }
      }
    }
  ]
}
```

Issue: parser fails with error `java.lang.RuntimeException: Unable to load RELATIVE ref: ./pet.json`. This happens because the `pet.json` is resolved RELATIVE to `parent.json` instead of `fancy_pet.json`. I expect parser needs to resolve `pet.json` w.r.t. to the origin i.e. `fancy_pet.json`

I have already found a fix for this issue and PR is on its way.

CC: @tzimisce012

gracekarina:
@EtonDoze Hi, is this related to https://github.com/swagger-api/swagger-parser/issues/567?

EtonDoze:
@gracekarina , its similar #604 but not same.
I created a PR #617 to fix this #604.

gracekarina:
@EtonDoze please check out the merge, to close the ticket

EtonDoze:
Closing it as its fixed via #617

PR #617, opened 2018-01-07
Fix for issue #616 allOf fails to resolve RELATIVE ref

EtonDoze:
Fix for Issue #616

Issue #616 arises due to an unchecked use-case where one external file is referenced with respect to the `parent` instead of the `source` where the reference is made.

Similar use-cases are handled correctly elsewhere but missed out in the case of `ComposedModels`. This issue is specific to ComposedModel only i.e. `allOf`. 

The fix leverages existing `join` function to normalize and resolve the paths before sending it for processing the external definitions.

CC: @tzimisce012 based on previous commits in this area.


gracekarina:
@EtonDoze  thanks!

Issue #628, opened 2018-01-15
OAS3 Resolving external schema containing allOf with $ref


joeljons:
When resolving a $ref to an external schema containing a $ref withing an allOf that internal $ref is not resolved.

The file to resolve: 
```
openapi: 3.0.0
info:
  version: 0.0.0
  title: oneOf and anyOf

paths:
  /oneOf:
    get:
      responses:
        '200':
          description: One of Cat or Dog
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: 'http://timotej.se/smartbear/pets.yaml#/components/schemas/Cat'
                  - $ref: 'http://timotej.se/smartbear/pets.yaml#/components/schemas/Dog'
  /anyOf:
    get:
      responses:
        '200':
          description: Any of Cat or Dog
          content:
            application/json:
              schema:
                anyOf:
                  - $ref: 'http://timotej.se/smartbear/pets.yaml#/components/schemas/Cat'
                  - $ref: 'http://timotej.se/smartbear/pets.yaml#/components/schemas/Dog'
```

Content of the external file:
```
openapi: 3.0.0
info:
  version: 0.0.0
  title: apa

components:
  schemas:
    Pet:
      required:
        - name
        - petType
      properties:
        name:
          type: string
        petType:
          type: string
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            huntingSkill:
              type: string
              default: lazy
              enum:
                - lazy
                - aggressive
          required:
            - huntingSkill
    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            packSize:
              description: The size of the pack the dog is from
              type: integer
          required:
            - packSize
```

Resolve using this code:
```
ParseOptions options = new ParseOptions();
options.setResolve(true);
OpenAPI openAPI = new OpenAPIV3Parser().readContents(yaml, null, options).getOpenAPI();
Yaml.pretty().writeValue(System.out, openAPI);
```

I would expect the output schema to contain Pet, Cat and Dog (since Pet is $ref:d from Cat and Dog).
The actual value is only Cat and Dog. The $ref to Pet is now broken.

I have tested this with the currently latest version of 2.0 branch (1292a4e1a77fd4fdc3c4b1dfe29b3d15ac88fa7b)

gracekarina:
[...]

PR #630, opened 2018-01-16
fixing allOf external resolving refs #628

gracekarina:
fixing allOf external resolving refs #628

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -8,6 +8,7 @@ import io.swagger.v3.oas.models.examples.Example;
 import io.swagger.v3.oas.models.headers.Header;
 import io.swagger.v3.oas.models.links.Link;
 import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.ComposedSchema;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.parameters.Parameter;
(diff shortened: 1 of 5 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
```diff
@@ -83,64 +84,88 @@ public String processRefToExternalSchema(String $ref, RefFormat refFormat) {
...
+            if(schema instanceof ComposedSchema){
+                ComposedSchema composedSchema = (ComposedSchema) schema;
+                if (composedSchema.getAllOf() != null){
+                    for(Schema item : composedSchema.getAllOf()){
+                        if (item.get$ref() != null){
+                            processRefSchema(item,file);
+                        }
+                    }
+
+                }else if (composedSchema.getOneOf() != null){
```
frantuma: replace `else if` with `if`, as allOf, anyOf and oneOf can theoretically coexist, check with a test case

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
```diff
@@ -83,64 +84,88 @@ public String processRefToExternalSchema(String $ref, RefFormat refFormat) {
...
+                    for(Schema item : composedSchema.getOneOf()){
+                        if (item.get$ref() != null){
+                            if (item.get$ref() != null){
+                                processRefSchema(item,file);
+                            }
+                        }
+                    }
+                }else if (composedSchema.getAnyOf() != null){
```
frantuma: replace `else if` with `if`, as allOf, anyOf and oneOf can theoretically coexist, check with a test case

PR #635, opened 2018-01-19
Fix/multiple composed model

gracekarina:
Fix/multiple composed model, depends on #630


gracekarina:
will  add more tests 👍

PR #636, opened 2018-01-22
Avoid overflow when storing an int64 in an Integer

joeljons:
In int64example.yaml the example value comes out as -81224344 in an Integer. 

Maybe there is a better way of checking for int64 examples, but this seems to do the job at least. I'm happy to change if there is some obvious better way this should be done in. :)

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -1193,7 +1193,9 @@ public class OpenAPIDeserializer {
             }
         }
         else if(v.getNodeType().equals(JsonNodeType.NUMBER)) {
-            value = v.intValue();
+            if (v.isInt()) {
+                value = v.intValue();
+            }
         }
         else if(!v.isValueNode()) {
             result.invalidType(location, key, "integer", node);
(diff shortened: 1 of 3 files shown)


gracekarina:
Hi @joeljons  I tried to push some changes but I have not permissions on your repo, so here is where I think it should be validated:

In the method `getInteger()`
like this:

```public Integer getInteger(String key, ObjectNode node, boolean required, String location, ParseResult result) {
        Integer value = null;
        JsonNode v = node.get(key);
        if (node == null || v == null) {
            if (required) {
                result.missing(location, key);
                result.invalid();
            }
        }
        else if(v.getNodeType().equals(JsonNodeType.NUMBER)) {
            if (v.isInt()) {
                value = v.intValue();
            }
        }
        else if(!v.isValueNode()) {
            result.invalidType(location, key, "integer", node);
        }
        return value;
    }

File: modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java [hunk 1]
```diff
@@ -1490,7 +1490,8 @@ public Object getAnyExample(String nodeKey,ObjectNode node, String location, Par
                 }
             } if (example.getNodeType().equals(JsonNodeType.NUMBER)) {
                 Integer integerExample = getInteger(nodeKey, node, false, location, result);
-                if (integerExample != null) {
+                JsonNode format = node.get("format");
```
gracekarina: Thanks!! @joeljons, I tried to push some changes but I have not permissions on your repo, so here is where I think it should be validated:

In the method getInteger()
like this (it would return null, and all integers will be validated as well):
line: 1196 : I added a boolean validation to make sure the number is an integer.

public Integer getInteger(String key, ObjectNode node, boolean required, String location, ParseResult result) {
        Integer value = null;
        JsonNode v = node.get(key);
        if (node == null || v == null) {
            if (required) {
                result.missing(location, key);
                result.invalid();
            }
        }
        else if(v.getNodeType().equals(JsonNodeType.NUMBER)) {
            if (v.isInt()) {
                value = v.intValue();
            }
        }
        else if(!v.isValueNode()) {
            result.invalidType(location, key, "integer", node);
        }
        return value;
    }

joeljons:
Your idea is much more elegant! :)
I will change my commit to implement it that way instead.

Issue #643, opened 2018-02-01
invalid duplicate references created w/ external file reference


jaz-ah:
I'm noticing a regression between 1.0.32 and 1.0.33 where a swagger spec defined below will create two XYZResponse objects instead of just one as in the 1.0.32 release - I believe this is a result of this change:
https://github.com/swagger-api/swagger-parser/commit/a5813748238448b14a61baf8ca732cee08561c88
@joeljons could you validate this issue - is there a different way I should be defining XYZResponse?  seems like this should be working ok:

main swagger file:
`{
    "swagger": "2.0",
    "info": {
        "version": "1.0.0",
        "title": "Sample Issue",
        "description": "API",
        "termsOfService": "TBD",
        "contact": {
            "name": "API Team"
        },
        "license": {
            "name": "MIT"
        }
    },
    "host": "xyz.com",
    "basePath": "",
    "schemes": [
        "https"
    ],
    "consumes": [
        "application/json"
    ],
    "produces": [
        "application/json"
    ],
    "paths": {
        "/xyz": {
            "$ref": "./xyz.json#/do_xyz"
	}
    },
    "definitions": {
	"XYZResponse": {
            "$ref": "./xyz.json#/definitions/XYZResponse"
	}
    }
}
`

xyz.json:
`
{
    "do_xyz": {
        "get": {
            "description": "do xyz stuff",
            "operationId": "doXyz",
            "produces": [
                         "application/json"
                         ],
            "parameters": [
            ],
            "responses": {
                "200": {
                },
                "default": {
                }
            }
        }
    },
    "definitions": {
        "XYZResponse": {
            "properties": {
		"foo": {
                    "type": "string"
		}
            }
        }
    }
}
`

jaz-ah:
hmm... although in this case @joeljons even though the source is different, the definition in xyz.json is never referenced - i.e. if I just generated based on xyz.json, XYZResponse wouldn't show up at all...

gracekarina:
Hi, So If I understand correctly, what we want is to avoid duplicated definitions, but also, in case there are two definitions named equally in different specs then both should appear in the main spec with `definition_number`, but only if the remote spec definition is referenced in the main one. what do you think, is that the issue? @joeljons  @jaz-ah

joeljons:
That is exactly how I see this.

jaz-ah:
yup that's right @gracekarina

gracekarina:
[...]

PR #651, opened 2018-02-22
fixing test migrated from master to 2.0 version of parser refs #643 also

gracekarina:
Migrating test from master version


gracekarina:
missing tests: RelativeReferenceTest.testIssue213

PR #660, opened 2018-03-20
Add logging to parser

mwhipple:


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
```diff
@@ -82,8 +86,13 @@ public OpenAPI read(String location, List<AuthorizationValue> auths, ParseOption
         OpenAPI output;
 
         List<SwaggerParserExtension> parserExtensions = getExtensions();
+	SwaggerParseResult parsed;
         for (SwaggerParserExtension extension : parserExtensions) {
-            output = extension.readLocation(location, auths,resolve).getOpenAPI();
+	    parsed = extension.readLocation(location, auths, resolve);
```
mwhipple: This bit seems the most important since the other classes collect messages which are subsequently dropped.

mwhipple:
The test failures seem unrelated but I can check them out later

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
gracekarina: I think is missing the .getOpenAPI();

mwhipple: That's below on L95, it was split out so the messages attached to the results could be output.

Issue #676, opened 2018-04-08
OpenAPI v2 converter - integer elements of enum are converted to String


ackintosh:
This issue reproduces in 2.0.0-rc3.

Given this Swagger Spec as input:

```json
{
    "swagger": "2.0",
    "info": {
        "version": "1.0",
        "title": "x-example"
    },
    "host": "httpbin.org",
    "basePath": "/anything",
    "schemes": [
        "http"
    ],
    "paths": {
        "/{foo}": {
            "get": {
                "parameters": [
                    {
                      "name": "enum_integer",
                      "in": "query",
                      "description": "",
                      "required": true,
                      "type": "integer",
                      "enum": [1, -2]
                    },
                    {
                      "name": "enum_integer_array",
                      "in": "query",
                      "description": "",
                      "required": true,
                      "type": "array",
                      "items": {
                          "type": "integer"
                      },
                      "collectionFormat": "csv",
                      "enum": [1, -2]
                    }
                ],
                "responses": {
                    "200": {
                        "description": "OK"
                    }
                }
            }
        }
    }
}
```

elements (`enum_integer` `enum_integer_array` in example) are converted to String as follows.

```java
    @Test(description = "OpenAPI v2 converter - integer elements of enum are converted to String")
    public void testIssue() throws Exception {
        OpenAPI oas = getConvertedOpenAPIFromJsonFile(ISSUE_JSON);
        List<Parameter> parameters = oas.getPaths().get(FOO_PATH).getGet().getParameters();

        List anEnum = parameters.get(0).getSchema().getEnum();

        System.out.println(anEnum.get(0).getClass().toString());
        // class java.lang.String

        System.out.println(anEnum.get(1).getClass().toString());
        // class java.lang.String
    }
```

I think it should be Integer. 💡

webron:
@ackintosh yes... and no :)

Your first example is valid and exhibits an issue.

Your second example is invalid. I mean, technically, it can be parsed, but the result would be that nothing could validate against it. You're saying it's an array of integers, but the enum is outside the definition and says that the value can only be this integer or that integer. The `enum` either has to be inside the `items`, or if it's meant to represent the array, then it should be `enum: [[1, -2]]`.

ackintosh:
Oh, I see. 💡 I understand that my second example is invalid.

PR #677, opened 2018-04-08
Fix for issue #676: convert integer elements of enum into the correct type

ackintosh:
This PR fixes issue #676 .


webron:
Thanks for the PR, @ackintosh! I'm sure @gracekarina will review it as soon as she can.

One point though - would it be ok to ask to add support for other types as well? It might as well work for boolean (I've seen it) and number as well.

ackintosh:
Thanks for the comment! I'll add support later.

PR #726, opened 2018-05-29
fixing relative references

gracekarina:
refs #727

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/RefUtils.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/RefUtils.java
@@ -54,7 +54,7 @@ public class RefUtils {
             result = RefFormat.URL;
         } else if(ref.startsWith("#/")) {
             result = RefFormat.INTERNAL;
-        } else if(ref.startsWith(".") || ref.startsWith("/")) {
+        } else if(ref.startsWith(".") || ref.startsWith("/") || ref.indexOf("#/") > 0) {
             result = RefFormat.RELATIVE;
         }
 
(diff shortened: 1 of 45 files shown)


JehandadK:
I have been looking into another issue, but wondering whether this can solve it. If we look into ResolverCache.java

```java
           contents = RefUtils.readExternalRef(file, refFormat, auths, parentDirectory);
            }
            else if(rootPath != null) {
                contents = RefUtils.readExternalUrlRef(file, refFormat, auths, rootPath);

```

It seems like rootPath is only used while resolving urls and hence I have to use different $ref paths to make it work for swagger-ui vs this parser.

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/RefUtils.java
```diff
@@ -136,27 +149,38 @@ public static String readExternalRef(String file, RefFormat refFormat, List<Auth
...
+        String result = null;
 
         try {
             if (refFormat == RefFormat.URL) {
-
                 result = RemoteUrl.urlToString(file, auths);
-
             } else {
                 //its assumed to be a relative file ref
                 final Path pathToUse = parentDirectory.resolve(file).normalize();
```
JehandadK: Should this parentDirectory be the directory of where the $ref exists or the directory of the main file which is read first?

I am asking this because this seems to be the behavior of swagger-ui. And to make my swagger spec work with both I had to use links.

gracekarina: is the directory where the file is

PR #753, opened 2018-07-06
Do not read Info block if it does not exist

bmordue:
Fixes issue described here: OpenAPITools/openapi-generator#177


ralphdoe:
Thanks a lot @jmini and @bmordue for the fix!

Issue #768, opened 2018-07-12
$ref in external file in requestBody with resolve option true


philippwinter:
Similar to Issue #749, JSON references inside the `requestBody`'s schema property are not properly resolved.

---

Main OAS 3.0 file:

```
openapi: 3.0.0
info:
  title: ref test
  version: '1.0'
servers:
  - url: 'http://localhost:8000/'
paths:
  /user/login:
    post:
      summary: Login
      operationId: loginUser
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "schemas.yaml#/User"
components:
  schemas: {}
```

Additional schema file:

```
User:
  type: "object"
  properties:
    Name:
      type: "string"
    Password:
      type: "string"
      format: "password"
```

Run this code:

        OpenAPIParser openApiParser = new OpenAPIParser();
        ParseOptions options = new ParseOptions();
        options.setResolve(true);

        String inputSpec = "main.yaml";

        OpenAPI openAPI = openApiParser.readLocation(inputSpec, null, options).getOpenAPI();
        String string = Yaml.mapper().writerWithDefaultPrettyPrinter().writeValueAsString(openAPI);

        System.out.println(string);

Current Value:

```
openapi: "3.0.0"
info:
  title: "ping test"
  version: "1.0"
servers:
- url: "http://localhost:8000/"
paths:
  /user/login:
    post:
      summary: "Login"
      operationId: "loginUser"
      requestBody:
        content:
          application/json:
            schema:
              $ref: "schemas.yaml#/User"
        required: true
components:
  schemas: {}
  extensions: {}
```

Expected Value:

```
openapi: "3.0.0"
info:
  title: "ping test"
  version: "1.0"
servers:
- url: "http://localhost:8000/"
paths:
  /user/login:
    post:
      summary: "Login"
      operationId: "loginUser"
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#components/schemas/User"
        required: true
components:
  schemas:
      User:
        type: "object"
        properties:
          Name:
            type: "string"
          Password:
            type: "string"
            format: "password"
  extensions: {}
```

Tested with version `2.0.0-SNAPSHOT` from Sonatype OSS

ymohdriz:
Hi @philippwinter 

I have raised a PR (https://github.com/swagger-api/swagger-parser/pull/769) for this issue.

Thanks,
Mohammed

ymohdriz:
Hi @philippwinter,

[...]

PR #769, opened 2018-07-12
Fix for issue 768 ($ref in external file in requestBody with resolve option true)

ymohdriz:
Hi team,

This PR is to solve the issue '$ref in external file in requestBody with resolve option true' as mentioned in #768 

Please review the PR and merge the same.

Thanks,
Mohammed


jmini:
This looks good to me. Thank you for your contribution and thank you for adding a test case to the PR.

gracekarina:
@ymohdriz  Hi, Why put '#/' as relative?

ymohdriz:
Hi @gracekarina,

If the reference starts with '#/' then it's internal.
If the reference starts with 'http', then it's url.
if the reference contains '#/' and also not starts with '#/', then it's a relative with the references file. For example, "$ref" : "reference.json#/definitions/sampleSchema" this is relative right?

Thanks,
Mohammed

ymohdriz:
Thanks @gracekarina for merging this commit.

Thanks,
Mohammed

Issue #805, opened 2018-08-20
problem with oas 3 that includes relative local file ref that includes remote file ref


fujigon:
### environment
- version 2.0.2
- Java test code
```java
public class OpenApiParserIT {

  @Test
  public void testParse() throws Exception {
    ParseOptions options = new ParseOptions();
    options.setResolve(true);
    options.setFlatten(true);

    OpenAPI openApi = new OpenAPIParser()
        .readLocation(Resources.getResource("parser/openapi_parser.yaml").toURI().toString(), null,
            options).getOpenAPI();
    System.err.println(openApi);
  }

}
```

- oas v3 file https://gist.github.com/fujigon/e2e1594a011022245e66fa6c21954ea9

### what I get
```
class OpenAPI {
    openapi: 3.0.0
    info: class Info {
        title: Sample APIs
        description: Sample API
        termsOfService: null
        contact: null
        license: null
        version: 1.0.0
    }
    externalDocs: null
    servers: [class Server {
        url: https://www.example.com/
        description: Server Url
        variables: null
    }]
    security: null
    tags: null
    paths: class Paths {
        {/tests/refToSchemaIncludingExternalRef/fromRef1=class PathItem {
            summary: null
            description: null
            get: class Operation {
                tags: null
                summary: null
                description: null
                externalDocs: null
                operationId: null
                parameters: null
                requestBody: null
                responses: class ApiResponses {
                    {default=class ApiResponse {
                        description: 200 response
                        headers: null
                        content: class Content {
                            {application/json=class MediaType {
                                schema: class Schema {
                                    title: null
                                    multipleOf: null
                                    maximum: null
                                    exclusiveMaximum: null
                                    minimum: null
                                    exclusiveMinimum: null
                                    maxLength: null
                                    minLength: null
                                    pattern: null
                                    maxItems: null
                                    minItems: null
                                    uniqueItems: null
                                    maxProperties: null
                                    minProperties: null
                                    required: null
                                    type: null
                                    not: null
                                    properties: null
                                    additionalProperties: null
                                    description: null
                                    format: null
                                    $ref: ./components/schemas/SchemaIncludingExternalRef.yaml#/components/schemas/SchemaIncludingExternalRef
                                    nullable: null
                                    readOnly: null
                                    writeOnly: null
                                    example: null
                                    externalDocs: null
                                    deprecated: null
                                    discriminator: null
                                    xml: null
                                }
                                examples: null
                                example: null
                                encoding: null
                            }}
[...]

PR #806, opened 2018-08-20
fix relative ref with http ref bug

fujigon:
This patch fixes https://github.com/swagger-api/swagger-parser/issues/805.

### DIFF
--- a/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/processors/ExternalRefProcessorTest.java
+++ b/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/processors/ExternalRefProcessorTest.java
@@ -6,9 +6,12 @@ import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.media.StringSchema;
 import io.swagger.v3.parser.ResolverCache;
+import io.swagger.v3.parser.core.models.AuthorizationValue;
 import io.swagger.v3.parser.models.RefFormat;
+import io.swagger.v3.parser.util.RemoteUrl;
 import mockit.Expectations;
 import mockit.Injectable;
+import mockit.Mocked;
 import mockit.StrictExpectations;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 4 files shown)


modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/processors/ExternalRefProcessorTest.java
```diff
@@ -138,4 +140,23 @@ public void testNestedExternalRefs(@Injectable final Schema mockedModel){
     	assertTrue(testedOpenAPI.getComponents().getSchemas().get("Contact") != null);
     	assertTrue(testedOpenAPI.getComponents().getSchemas().get("Address") != null);
     }
+
+	@Test
+	public void testRelativeRefIncludingUrlRef(@Injectable final Schema mockedModel) {
```
jmini: This requires a server call to `https://raw.githubusercontent.com/`. Please mock it in the unit tests.

jmini:
The change looks good to me, but your new unit test requires a server call to `https://raw.githubusercontent.com/`. Please mock it in the unit tests.

Add a new parameter: `@Mocked RemoteUrl remoteUrl` and the expectations at the beginning of your method:

```java
	        final String url = "https://raw.githubusercontent.com/swagger-api/swagger-parser/v2.0.2/modules/swagger-parser-v3/src/test/resources/relative/globals.yaml";
	        final String expectedResult = "components:\n" + 
	                "  schemas:\n" + 
	                "    link-object:\n" + 
	                "      type: object\n" + 
	                "      additionalProperties:\n" + 
	                "        \"$ref\": \"#/components/schemas/rel-data\"\n" + 
	                "    rel-data:\n" + 
	                "      type: object\n" + 
	                "      required:\n" + 
	                "      - href\n" + 
	                "      properties:\n" + 
	                "        href:\n" + 
	                "          type: string\n" + 
	                "        note:\n" + 
	                "          type: string\n" + 
	                "    result:\n" + 
	                "      type: object\n" + 
	                "      properties:\n" + 
	                "        name:\n" + 
	                "          type: string\n" + 
	                "        _links:\n" + 
	                "          \"$ref\": \"#/components/schemas/link-object\"\n" + 
	                "";
	        List<AuthorizationValue> auths = null;

	        new Expectations() {{
                    RemoteUrl.urlToString(url, auths);
	            times = 1;
	            result = expectedResult;
	        }};
```

---

Second question:
Is is use that the old behavior adding `./` (I guess that this was for relative url) is still present for relative url?

fujigon:
Thanks for the review, and fine example to support this PR.
I reflected your review point.

fujigon:
I added a test to confirm that "./" is still attached to relative local ref file.

jmini:
I am not owner of this project, but this PR looks good to me.

Thank you a lot for this contribution

gracekarina:
Thanks for this PR @fujigon

douglasbgray:
@gracekarina I ran into this issue in the 2.0.2 release and was about to file an issue, when I see it has been fixed here. Do you have a timeline for the 2.0.3 release with this fix?

Thanks!

PR #809, opened 2018-08-22
Update Readme to use badges from Shields.io

pavog:
Since issuestats.com seems to stay offline we should use shields as it's the best alternative afaik.


pavog:
Jenkins errors because the files for the testresults haven't been cleaned/renewed in the pr build workspace:

```
hudson.AbortException: Test reports were found but none of them are new. Did leafNodes run? 
For example, /var/lib/jenkins/workspace/oss-swagger-parser-master-pr/modules/swagger-compat-spec-parser/target/surefire-reports/TEST-TestSuite.xml is 11 hr old
```

We should consider cleaning by removing all test resultfiles.
Or simply renew/touch the files with `touch *.xml` in the results dir.
Or maybe we can skip creating result files with `mvn test -Dsurefire.useFile=false` or setting it in the pom.xml:

```
<plugin>
  <artifactId>maven-surefire-plugin</artifactId>
  <version>2.10</version>
  <configuration>
    <useFile>false</useFile>
  </configuration>
</plugin>
```

Let my know which solution you think fits best.
I can file a pr then.

gracekarina:
Hi @pavog can you please update this PR, with the current release? thanks!

gracekarina:
Thanks! @pavog

PR #815, opened 2018-08-24
Support for empty string as default schema value

### LINKED ISSUE #814: Empty default in schema is parsed as null
#### Parser version:
2.0.2
#### Description:
OAS 3 spec (partial): 
```yaml
components:
  parameters:
    example:
      name: example
      description: example
      in: query
      required: false
      schema:
        type: array
        items:
          type: string
        default: ''
```

The parameter schema parser treats empty strings as null.

`schema.getDefault()` returns `null`.

The following results in a proper default value being set:
```yaml
components:
  parameters:
    example:
      name: example
      description: example
      in: query
      required: false
      schema:
        type: array
        items:
          type: string
        default: 'a,b,c'
```

`schema.getDefault()` returns `"a,b,c"`.

jason-cohen:
Fix #814

`isBlank` resulted in the default value not being parsed, when the field was in fact explicitly set.

Now we would only consider null as the absence of a field in the specification.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -2125,7 +2125,7 @@ public class OpenAPIDeserializer {
         }
 
         value = getString("default", node, false, location, result);
-        if (StringUtils.isNotBlank(value)) {
+        if (value != null) {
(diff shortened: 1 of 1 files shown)


JonathanParrilla:
please build

JonathanParrilla:
please build

gracekarina:
Hi, the scope of the issue changed, the validation of this new scope made in another PR. Thanks @jason-cohen for reporting this and sorry for the long reply. I will close this because is no longer an issue.

PR #829, opened 2018-09-05
prepare release 2.0.3

gracekarina:


frantuma:
please build

frantuma:
please build

pom.xml
```diff
@@ -253,7 +253,7 @@
         <swagger-parser-v2-version>1.0.38-SNAPSHOT</swagger-parser-v2-version>
```
jmini: Having a snapshot in the dependencies is not a good idea in my opinion. Release should only have released versions in their dependencies. 

Otherwise, builds can break (if the `SNAPSHOT` version is remove or if it evolves to something that is no longer compatible)

frantuma:
@jmini Thanks for spotting this! just released `2.0.4` correcting the issue.

Issue #837, opened 2018-09-12
Examples with Ref not parsed when in responses


JehandadK:
Simply put

```yaml
responses:
  content:
    examples:
       key:
         $ref: 'file.yaml#components/examples/key'

```

Does not get resolved.

ymohdriz:
Hi @JehandadK 

Can you please provide a sample open api file to reproduce this issue?

Thanks,
Mohammed Rizwan

gracekarina:
Fixed by #838

PR #838, opened 2018-09-12
Issue 837 oas3 

JehandadK:
To demonstrate and confirm whether this is a valid issue. I have created a test case to show what my parser is expecting. But it simply seems like the Examples are not parsed in responses.
As in #837


JehandadK:
@gracekarina Btw, which branch should I merge in?

gracekarina:
Hi what do you mean?

PR #852, opened 2018-10-01
Fix NullPointerException when resolving an ApiResponse with no content

tjaneczko:
This PR fixes a NullPointerException which occurs when an `ApiResponse` is read from an external reference which does not have any content defined.


tjaneczko:
Will do!  Thanks for the prompt review!

tjaneczko:
@gracekarina Added a test for loading an external response with no content.  Also looks like the `DeserializationUtils` class was being completely mocked in these tests, meaning the existing `testLoadExternalRefWithEscapedCharacters` was never completely run, so I fixed that up and changed it to a partial mock.

Issue #434, opened 2017-04-05
Some $ref values from external files are not updated with the processed internal reference values


leochr:
In some cases, the reference values coming from external files are not updated with the processed internal reference value.

The value returned by invoking io.swagger.parser.processors.ExternalRefProcessor.processRefToExternalDefinition(String, RefFormat) method should be set on the reference. It's done for one instance, but not for others.

                if (isAnExternalRefFormat(refModel.getRefFormat())) {
                    refModel.set$ref(processRefToExternalDefinition(refModel.get$ref(), refModel.getRefFormat()));
                } else {
                    **processRefToExternalDefinition(file + refModel.get$ref(), RefFormat.RELATIVE);**
                }

Above method invocation in bold should be 
refModel.set$ref(processRefToExternalDefinition(file + refModel.get$ref(), RefFormat.RELATIVE));

Similar behaviour in io.swagger.parser.processors.ExternalRefProcessor.processRefProperty(RefProperty, String) method. 

    private void processRefProperty(RefProperty subRef, String externalFile) {
        if (isAnExternalRefFormat(subRef.getRefFormat())) {
            String $ref = constructRef(subRef, externalFile);
            subRef.set$ref($ref);
            if($ref.startsWith("."))
                processRefToExternalDefinition($ref, RefFormat.RELATIVE);
            else {
                processRefToExternalDefinition($ref, RefFormat.URL);
            }

        } else {
            **processRefToExternalDefinition(externalFile + subRef.get$ref(), RefFormat.RELATIVE);**
        }
    }

Above method invocation in bold should be 
subRef.set$ref(processRefToExternalDefinition(externalFile + subRef.get$ref(), RefFormat.RELATIVE));

In above method, for external ref formats, the reference value is updated correctly because later on in DefinitionsProcessor the definitions are processed one by one. At that time the external RefProperty value will be replaced by internal value. But the same external reference values is processed twice unnecessarily by processRefToExternalDefinition. Setting the internal value returned by processRefToExternalDefinition on the RefProperty the first time should avoid this and doing so would keep the behaviour consistent with the other invocations of processRefToExternalDefinition.

leochr:
I am able to demonstrate this problem by slight modifications to existing test cases. I'll create a PR to fix this issue and will include the test cases as well.

gracekarina:
fixed by #864

PR #439, opened 2017-04-06
Fixes for ExternalRefProcessor issues #434 and #435

leochr:
Fixes 
- #434 - Some $ref values from external files are not updated with the processed internal reference values 
- #435 - Relative reference can also start with "/"

Issue #435, opened 2017-04-05
Relative reference can also start with "/"


leochr:
In io.swagger.parser.processors.ExternalRefProcessor.processRefProperty(RefProperty, String), checking whether the relative reference starts with "."

Relative reference can also start with "/" (join method within the same class, check correctly)

    private void processRefProperty(RefProperty subRef, String externalFile) {
        if (isAnExternalRefFormat(subRef.getRefFormat())) {
            String $ref = constructRef(subRef, externalFile);
            subRef.set$ref($ref);
            if($ref.startsWith("."))
                processRefToExternalDefinition($ref, RefFormat.RELATIVE);
            else {
                processRefToExternalDefinition($ref, RefFormat.URL);
            }

        } else {
            processRefToExternalDefinition(externalFile + subRef.get$ref(), RefFormat.RELATIVE);
        }
    }

leochr:
I'll create a PR to fix this issue.

gracekarina:
Closed by #864

PR #864, opened 2018-10-05
Fixes for issue #435 and #434 because the original PRs are stale

JehandadK:
related to #439


JehandadK:
@gracekarina I am unsure of whether issue #434 has correct expectations. I have put the Unit Tests, can you tell me whether they make sense in the second commit? I am unfamiliar with swagger 2. I have only used swagger 3.

gracekarina:
@JehandadK  yes they make sense thanks! So this PR fixes #434 and add a test for #435 right?

PR #888, opened 2018-10-19
 Fix and Unit test for Issue 887

r-sreesaran:
Duplicate Tags are not getting removed in swagger parser

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -97,6 +97,8 @@ public class OpenAPIDeserializer {
 
 
 
+    private final Set<String> operationIDs = new HashSet<>();
+
     public SwaggerParseResult deserialize(JsonNode rootNode) {
     	return deserialize(rootNode, null);
     }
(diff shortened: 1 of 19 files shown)


r-sreesaran:
adapted code as per requested changes

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -297,20 +297,23 @@ public Components getComponents(ObjectNode obj, String location, ParseResult res
...
         }
-        
-//        Map<String, Tag> tags = new LinkedHashMap<String,Tag>();
         List<Tag> tags = new ArrayList<>();
         for (JsonNode item : obj) {
             if (item.getNodeType().equals(JsonNodeType.OBJECT)) {
                 Tag tag = getTag((ObjectNode) item, location, result);
-                if (tag != null ) {
-                    if(tags.contains(tag.getName())) {
-                        LOGGER.warn("Duplicate tags found: "+tag.getName());
```
gracekarina: Hi, The SwaggerParseResult Object has  a map of messages: https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-core/src/main/java/io/swagger/v3/parser/core/models/SwaggerParseResult.java#L7

The main idea is to use that to warn the user, maybe look the way this PR implemented something similar:
https://github.com/swagger-api/swagger-parser/pull/873/commits/15a3881540ade50cd1c364aef8645491f046e6dd#diff-998f064dc6ef209e504969ef62654a84R2695

modules/swagger-parser/src/test/java/io/swagger/parser/OpenAPIParserTest.java
```diff
@@ -67,7 +67,6 @@ public void testSimple() {
 
     @Test
     public void testDuplicationOfTags() {
```
gracekarina: If you implement the messages map as mentioned before, please add the validation of the message here.

Issue #895, opened 2018-10-28
Swagger parser fails to validate valid url is given in License information


r-sreesaran:
Swagger parser fails to validate valid url is given in License information. 

By Specification 
"
A URL to the license used for the API. MUST be in the format of a URL.
"

This is not getting validated as of now.

r-sreesaran:
The same is the case for url in contact object.

r-sreesaran:
I have fixed the issue and also raise a pr for the issue

PR #896, opened 2018-10-28
Fix for issue 895

r-sreesaran:
Fix for issue #895


r-sreesaran:
@gracekarina kindly do let me know if there is anything else that needs to be done from my side

PR #902, opened 2018-11-02
Fix Serialization of Enums

Xyaren:
Enums were serialized uppercase instead of their encapsuled string "value"


gracekarina:
Hi @Xyaren can you please add test for this change? Thanks

gracekarina:
Hi @Xyaren can you please add test for this change? Thanks

PR #914, opened 2018-11-06
Recursion and resolveFully can give infinite loop while serializing

joeljons:
I'm not sure if the problem is in resolveFully or in swagger-core for the `Json.mapper().writeValueAsString(openAPI)` part.

This PR contains just a failing test pointing at the problem.


gracekarina:
fixed by #943

Issue #913, opened 2018-11-06
Unable to load RELATIVE $ref property


eugeniace:
We defined a swagger file ( src/main/resources/BS/ApiSpecification.swagger ) that references a property defined in a JSON file (src/main/resources/BO/Resource/ApiSpecificationBO.json). 
In ApiSpecificationBO.json file there is a relative $ref property defined in another JSON (src/main/resources/BO/Common/BasicComponents.json). 

The class generating fails with the fallowing error from swagger-parser:
Unable to load RELATIVE ref: ../Common/BasicComponents.json path: C:\utils-app\src\main\resources\BS

As we can see, it tries to resolve the reference to the second JSON file considering as a parent folder the one where the initial swagger file is placed, and not the folder where the JSON file containing the $ref is placed.

We used version v1.0.39.

Resources:

----------------------------------------------
src/main/resources/BS/ApiSpecification.swagger:
----------------------------------------------
```
[...]
"responses": {
                    "200": {
                        "schema": {
                            "type": "array",
                            "items": {
                                "$ref": "#/definitions/accessProfileBO"
                            }
                        }
                    }
                }
[...]
"definitions": {
        "accessProfileBO": {
            "$ref": "../BO/Resource/ApiSpecificationBO.json"
        }
    },
[...]
```
------------------------------------------------------
src/main/resources/BO/Resource/ApiSpecificationBO.json
------------------------------------------------------
```
{
    "properties": {
        "details": {
            "type": "object",
            "properties": {
                "confirmationCodeRequired": {
                    "$ref": "../Common/BasicComponents.json#/properties/indicatorType"
                }
            }
        }
    }
}
```
------------------------------------------------------
src/main/resources/BO/Common/BasicComponents.json
------------------------------------------------------
```
{
    "properties": {
        "indicatorType": {
            "type": "object",
            "properties": {
                "indicator": {
                    "type": "boolean"
                }
            }
        }
    }
}
```

[src.zip](https://github.com/swagger-api/swagger-parser/files/2552463/src.zip)

eugeniace:
Please approve the fix for this issue:
https://github.com/swagger-api/swagger-parser/pull/917

gracekarina:
Fixed by #917  please let us know if this keeps happening. Thanks!

PR #917, opened 2018-11-08
Fix for #913 - problem with properties in object #ref

eugeniace:
Please approve the fix for #913


gracekarina:
@eugeniace thanks!

PR #920, opened 2018-11-10
Fix consider ResponseSchema refModel's format

VIIgit:
as it can differ e.g. INTERNAL, RELATIVE

The provided unit test example did not work before, as a reference could not be resolved.


gracekarina:
Hi @VIIgit I think what was failing was the branch, now `master` branch is OAS 3 support, and `v1` branch is 1.x support. If there is something else you wanted to point out, let us know.

gracekarina:
moved to #921

PR #921, opened 2018-11-11
fix Consider refModel's RefFormat then provided refFormat

VIIgit:


gracekarina:
I'm sorry @VIIgit but I still don`t understand the issue. Can you please explain. thanks.

VIIgit:
Without the fix, not all references were resolved:

Parser's Message returns:
```
[ "Invalid ref format: ./sub-folder/common-responses.yaml#/responses/Ok201#/definitions/User" ]
```

By resolving a reference which is referring again a relative linked file then the format "INTERNAL" is used to resolve the reference instead of correct refModel's format "RELATIVE".

Without the fix, the provided swagger-root.yaml example doesn't resolve response 201
```
{
  "swagger" : "2.0",
  ...
  "paths" : {
    "/users" : {
      "get" : {
        "summary" : "Returns a list of users.",
        "description" : "Optional extended description in Markdown.",
        "produces" : [ "application/json" ],
        "parameters" : [ ],
        "responses" : {
          "200" : {
            "description" : "OK",
            "schema" : {
              "$ref" : "#/definitions/User"
            }
          },
          "201" : {
            "$ref" : "./sub-folder/common-responses.yaml#/responses/Ok201"
          }
        }
      }
    }
  },
...
```

after the fix the all references are resolved correctly
```
{
  "swagger" : "2.0",
...
  "paths" : {
    "/users" : {
      "get" : {
        "summary" : "Returns a list of users.",
        "description" : "Optional extended description in Markdown.",
        "produces" : [ "application/json" ],
        "parameters" : [ ],
        "responses" : {
          "200" : {
            "description" : "OK",
            "schema" : {
              "$ref" : "#/definitions/User"
            }
          },
          "201" : {
            "description" : "OK",
            "schema" : {
              "$ref" : "#/definitions/User"
            }
          }
        }
      }
    }
  },
...
```

john3300:
Will/can this be added to master too?

gracekarina:
Please send a PR, will make it faster 👍

VIIgit:
> Will/can this be added to master too?

The exact same data constellation but as OpenApi 3 files and with openApiParser works as expected. So I believe there is no need to add it to the master.

john3300:
Fair enough, thanks @VIIgit.

I did check out this branch and tried to run the tests that you added with Maven but none of the tests in that class seem to be executed.  Am I missing something?

john3300:
Rock on, thanks!

Issue #922, opened 2018-11-11
Deprecated field in Schemas Object must be set to false by default


r-sreesaran:
By specification :
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaObject

"Specifies that a schema is deprecated and SHOULD be transitioned out of usage. default value is false"

The default value needs to be set as false for all schema, but currently, it is not initialized.

r-sreesaran:
I have raised a PR for the issue.

gracekarina:
Closed By #923

frantuma:
@r-sreesaran #950 reverts #933 as according to the spec, `deprecated` is an optional property, which defaults to `false` when not specified; this doesn't mean that it must be explicitly set to `false`, and I would say it's not desired to have the parser explicitly set to their default values all optional properties with a default value in spec (at least not as default, will open a new ticket to activate this behaviour with options flag); please let me know if you still think we should maintain this behaviour as standard instead of activated via an option.
The main drawback of the reverted behaviour is the "pollution" of the spec e.g with mass `deprecated = false` fields, also it slightly changes the original spec (non present properties -> properties with default values) which again I don't feel should be the default behaviour

r-sreesaran:
@frantuma understood your point, that seems to be valid. Thanks for reverting the changes.

PR #923, opened 2018-11-11
Fix for issue 922

r-sreesaran:
Fix for issue #922


gracekarina:
@r-sreesaran thanks! @ymohdriz thanks for your comments

PR #929, opened 2018-11-14
Recursion that can lead to an infinite loop in ResolverFully

joeljons:
This is another case of infinite recursion. In contrast to #914 this fails already in ResolveFully.

This PR contains just a failing test pointing at the problem.


gracekarina:
Fix by #943

Issue #941, opened 2018-11-20
OpenAPIV3Parser.read: UTF-8 Charset errors


Rush0001:
Hello,
parsing a OAS 3.0.0 spec in UTF-8 with this configuration :
`openapi: 3.0.0`
`info:`
`  title: ApiWeb`
`  version: 1.0.0`
`  description: "Documentation de l'API élaboré par nos soins"`

displays wrong characters:
Documentation de l'API **??**labor**??** par nos soins

The yaml source file is in UTF-8. (i checked in intelliJ and in another editor)
OpenAPIV3Parser.read doesn't seem to support correctly UTF-8.

Thanks for your help

ymohdriz:
Hi @Rush0001,

We tried to reproduce the issue, it's not reproducible and newly added tests are passing.
Could you please attach the input openapi yaml file which you used?
Which parser version are you using?

Thanks,
Mohammed

Rush0001:
Hello @ymohdriz ,

This is the yaml file and a screenshot.

Thanks for your help.
Rush0001

[file.txt](https://github.com/swagger-api/swagger-parser/files/2611465/file.txt)
![screenshot](https://user-images.githubusercontent.com/16193874/48956407-219c1000-ef53-11e8-96c4-1c1990b83087.png)

gracekarina:
Hi @Rush0001  I added a test with the file you sent, but I haven't been able to reproduce, because the test passes.  please check the commit above.

ymohdriz:
Hi @Rush0001,

Looks like the issue is in [swagger-ui](https://github.com/swagger-api/swagger-ui). Can you raise an issue in swagger-ui.

Thanks,
Mohammed

jsPl:
Could be due to the operating system default encoding other than UTF-8.
Try running your JVM with -Dfile.encoding=UTF8 to override defaults.

PR #942, opened 2018-11-21
Test added for issue 941

ymohdriz:
Hi team,

This test indicates that the issue mentioned in #941 is not reproduced.

Thanks,
Mohammed


gracekarina:
thanks @ymohdriz

Issue #959, opened 2018-12-01
Parser Fails to log a warning when parameters are duplicated


r-sreesaran:
According to the specification. 
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#path-item-object

The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a name and location. The list can use the Reference Object to link to parameters that are defined at the OpenAPI Object's components/parameters.

But when a parameter is duplicated, the parser fails to throw a warning message.

gracekarina:
closed by #960

PR #960, opened 2018-12-01
added fix for issue 959

r-sreesaran:
fix for issue #959


gracekarina:
Hi, thanks for this PR. I suggest to divide the test in two, instead of evaluating two different specs in one test. @r-sreesaran

Issue #966, opened 2018-12-18
Preserve body parameter name via the vendor extension


wing328:
Consider the following OAS v2 spec, the body parameter name "body" is completely lost after parsing the spec:
```
      parameters:
        - in: body
          name: body
          description: Pet object that needs to be added to the store
          required: true
          schema:
            $ref: '#/definitions/Pet'
```

To preserve the body parameter name, we can store it in the vendor extension `x-codegen-request-body-name`

blast-hardcheese:
Just ran into this myself. Structurally, it's immaterial, as OpenAPI 3.x+ doesn't have names for body parameters. That being said, it does not offer a pleasant experience for users migrating to the new structure using the swagger converter module.

gracekarina:
closed by #967

Issue #748, opened 2018-07-03
OpenAPI v2 converter: parameter name is lost


jmini:
When the parser is doing the conversion from OAS2 to OAS3, a parameter of type `in: "body"` in OAS2 is converted to a `requestBody` in OAS3.

Example:

OAS2:

```yaml
      - name: "requestList"
        in: "body"
        description: "Requests to post"
        required: true
        schema:
          $ref: "#/definitions/listitems"
```

OAS3:

```yaml
      requestBody:
        description: Requests to post
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/listitems'
        required: true
```

During the operation, the parameter name is lost. Would it be possible to have a `x-name` added to the request body, in order to keep the information of the OAS2?

Expected result after the change?

```yaml
      requestBody:
        x-name: "requestList"
        description: Requests to post
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/listitems'
        required: true
```

If you are interested I might be able to propose a PR for that.
Thank you in advance for your answer.

---

Complete input example:

```yaml
swagger: "2.0"
info:
  description: Container Application Model Description
  version: 1.0.0
  title: ContainerApplication
  
host: localhost:32000
basePath: /api/v1.0
schemes:
- http

tags:  
- name: ApplicationEndpointTest
  x-displayName: "ApplicationEndpointTest"
  description: |
    "Test of endpoints "
    
paths:
  /default/EndpointTest: 
    post:
      tags:
      - ApplicationEndpointTest      
      description: "post a request"
      operationId: callEndpoint
      consumes:
      - application/json
      parameters:
      - name: "requestList"
        in: "body"
        description: "Requests to post"
        required: true
        schema:
          $ref: "#/definitions/listitems"
      responses:
        200:
          description: "Successful response"
        404:
          description: "The request was not successfully executed."
      
  
definitions:
  listitems : 
    description: "list of items"
    type: array
    items: 
      type: "string"
```

PR #967, opened 2018-12-18
Ensure the body parameter name is preserved

wing328:
For https://github.com/swagger-api/swagger-parser/issues/966

Related: https://github.com/swagger-api/swagger-parser/issues/748


jmini:
Is there any feedback on this PR?

Are you rejecting the proposition?
Or have you any concerns about this?

A while ago I did open a similar issue about preserving the name of the body parameter from the v2 Spec as extension in v3: #748
Back then I proposed an other approach (an other name, an other position of the extension), but the idea is the same and for me this would also be OK.

Please let us know if we can do anything to help this feature to moving forward (this PR can be rewritten if an other design decision is considered)

advance512:
Please move forward with this. It has been a month and still no response - this causes major issues in dependent projects. Thank you

JonathanParrilla:
please build

modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
```diff
@@ -561,6 +562,7 @@ public Operation convert(io.swagger.models.Operation v2Operation) {
                     formParams.add(param);
                 } else if ("body".equals(param.getIn())) {
                     operation.setRequestBody(convertParameterToRequestBody(param, v2Operation.getConsumes()));
+                    operation.addExtension("x-codegen-request-body-name", param.getName());
```
blast-hardcheese: It seems as though the `name` field's intent has been replaced by `title`, as evidenced by how the string `"body"` gets resolved in the converter: https://github.com/swagger-api/swagger-parser/blob/670b278f1cf925ce93a8d3c17a1c8411d9e5329b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/InlineModelResolver.java#L67

I would propose a different direction, instead of adding a vendor extension, to just `setTitle(param.getName())` and let the rest of the infrastructure wire the parameter through.

This will seamlessly resolve the issue for those who have migrated to the new `title` infrastructure.

blast-hardcheese: @gracekarina Out of curiosity, any thoughts on the direction of this comment? I still suspect adding a vendor extension that duplicates in-built behavior is causing a needless fork in functionality, but maybe I've misunderstood an aspect about the semantics between `name` and `title` here.

gracekarina: Hi @blast-hardcheese I'll review the comment you made, if what you say is right I'll revert the PR. thanks for pointing us to that.

DrSatyr:
@wing328 could you please describe the purposes of this "x-codegen-request-body-name" extension. I am trying to figure out it and decide is it needed in my case or not. Am I right that it's needed for code generation in some languages?

Issue #976, opened 2019-01-09
validation doesn't recognize reference in path parameter


fujigon:
### Test
test code
```
    @Test
    public void testValidationIssue() {
        ParseOptions parseOptions = new ParseOptions();
        parseOptions.setResolveFully(true);
        SwaggerParseResult result = new OpenAPIV3Parser().readLocation("src/test/resources/path-parameter-validation.yaml", null, parseOptions);
        System.err.println(result.getMessages());
        assertThat(result.getMessages().size(), CoreMatchers.is(0));
    }
```

`path-parameter-validation.yaml`

```
openapi: 3.0.0
info:
  title: Sample APIs
  version: 1.0.0
  description:
    Sample API

components:
  parameters:
    accountId:
      name: accountId
      schema:
        type: string
      in: path
      required: true

paths:
  /users/{accountId}:
    get:
      operationId: ReadUser
      parameters:
      - $ref: '#/components/parameters/accountId'
      responses:
        '200':
          description: 200 response
```

then the result is...

```
[attribute paths.'/users/{accountId}'. Declared path parameter accountId needs to be defined as a path parameter in path or operation level]

java.lang.AssertionError: 
Expected: is <0>
     got: <1>
```

gracekarina:
fixed by #977

Issue #911, opened 2018-11-05
Not able to identify the missed path variables


parameshjava:
Hi,

It's showing a valid JSON though the path variables are missed to declare. We can find the error, if we keep the same JSON in editor.swagger.io.

Please find the below sample JSON:

```
{
  "swagger": "2.0",
  "info": {
    "description": "Pets Reference Data",
    "version": "1.0",
    "title": "Pets Reference Data"
  },
  "host": "localhost:8080",
  "basePath": "/api/pets/referencedata/v1.0",
  "tags": [
    {
      "name": "Pets Reference Data Operations",
      "description": "Pets Reference Data Controller"
    }
  ],
  "paths": {
    "/{tenantId}/categories/{categoryCode}": {
      "get": {
        "tags": [
          "Dynamic Codces Operations"
        ],
        "summary": "Reference Data Service",
        "description": "Retrieves the ReferenceData for a given tenant id, category and sub category with language code as english",
        "operationId": "Reference Data Service_2",
        "produces": [
          "*/*"
        ],
        "parameters": [
          {
            "name": "categoryCode",
            "in": "path",
            "description": "categoryCode",
            "required": true,
            "type": "string"
          },
          {
            "name": "subCategoryCode",
            "in": "query",
            "description": "subCategoryCode",
            "required": false,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "OK",
            "schema": {
              "$ref": "#/definitions/petsDataResponse"
            }
          },
          "401": {
            "description": "Unauthorized"
          },
          "403": {
            "description": "Forbidden"
          },
          "404": {
            "description": "Not Found"
          }
        }
      }
    }
  },
  "definitions": {
    "Domain": {
      "type": "object",
      "properties": {
        "b2X": {
          "type": "string"
        },
        "documentation": {
          "type": "string"
        },
        "key": {
          "type": "string"
        },
        "memberValue": {
          "type": "string"
        },
        "name": {
          "type": "string"
        },
        "verbalization": {
          "type": "string"
        }
      },
      "title": "Domain"
    },
    "Error": {
      "type": "object",
      "properties": {
        "errorCode": {
          "type": "string"
        },
        "errorDescription": {
          "type": "string"
        }
      },
      "title": "Error"
    },
    "petsDataResponse": {
      "type": "object",
      "properties": {
        "messageElements": {
          "$ref": "#/definitions/messageElements"
        },
        "tenant": {
[...]

PR #977, opened 2019-01-09
fix bug of validation failure when path parameter contains ref

fujigon:
This patch fixes ~https://github.com/swagger-api/swagger-parser/issues/911~ https://github.com/swagger-api/swagger-parser/issues/976


jjlin:
Did you mean to say it fixes #976?

fujigon:
exactly.
Thanks for following up!

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -558,7 +558,7 @@ private boolean isPathParamDefined(String pathParam, List<Parameter> parameters)
             return false;
         } else {
             Parameter pathParamDefined = parameters.stream()
-                            .filter(parameter -> pathParam.equals(parameter.getName()) && "path".equals(parameter.getIn()))
+                            .filter(parameter -> (parameter.get$ref() != null) || (pathParam.equals(parameter.getName()) && "path".equals(parameter.getIn())))
```
mih-kopylov: This checks only having $ref, but does not check that the reference is for this name and contains `in: path`

mih-kopylov: Actually, yep, I agree to fix false matching for valid file even with workaround.

From my pov this kind of validation should be called outside of deserialization, even after resolving all $refs, and only in case when ParseOptions have $refs resolving enabled.
Somewhere here, probably: https://github.com/swagger-api/swagger-parser/blob/77f69e9407890c472f5b7c0b5c2c34d484dac4e5/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java#L65

What do you think?

fujigon: Agree.
The root problem is that currently validation and deserialization is tightly coupled and ref-resolve is not considered.
We should de-couple them. validation should be done **after** deserialization.

I quickly read OpenAPIDeserializer and found that many validation is done in it 😢 (easily find with test searching `result.warning(`). 
It seems to cost much work to separate deserialization and validation...
So I think we should create an issue about the decoupling, work on it in the future.

Let me know what you think. :)

modules/swagger-parser-v3/src/test/resources/path-parameter-validation.yaml
```diff
@@ -0,0 +1,25 @@
...
+  description:
+    Sample API
+
+components:
+  parameters:
+    accountId:
+      name: accountId
+      schema:
+        type: string
+      in: path
```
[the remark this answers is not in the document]
fujigon: added https://github.com/swagger-api/swagger-parser/pull/977/files#diff-e3d6dbd47e8546ab4b5f4fa5fd87146aR1654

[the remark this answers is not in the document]
mih-kopylov: Great, thanks! 
But now this test doesn't check that case and it doesn't fail. 
Will there be a separate ticket to write an implementation of this test? 
It can be simply forgotten after some time either, I guess.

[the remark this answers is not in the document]
fujigon: https://github.com/swagger-api/swagger-parser/issues/982
issued!

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
fujigon: https://github.com/swagger-api/swagger-parser/issues/982
issued!

mih-kopylov:
@gracekarina could you please take a look at this PR? It might be useful

fantavlik:
I would love to get this reviewed! My team is currently blocked from using swagger-parser due to this bug. Let me know if there's anything I can do to help!

gracekarina:
looking into it, sorry for the delay.

modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java
```diff
@@ -1759,7 +1759,6 @@ public void testValidationIssueFalseNegative() {
         ParseOptions parseOptions = new ParseOptions();
         parseOptions.setResolveFully(true);
         SwaggerParseResult result = new OpenAPIV3Parser().readLocation("src/test/resources/validation/path-parameter-validation-ref-invalid.yaml", null, parseOptions);
-
         // should detect, but validation misses the error of the content in the ref, in order to avoid false positive
```
gracekarina: Hi, what is this test doing? there are no assertion. please remove the comments.

fujigon:
Thanks for your review.
I reflected your point!
Could you kindly review this again please?

Thanks!

fujigon:
@gracekarina :)

drewish:
It seems like this bug is still present in the v1 branch used by the validator-badge https://github.com/swagger-api/validator-badge/issues/137

JFCote:
@gracekarina This bug it still present in the v1 branch.

Is it possible to:
1) Apply this PR to the `v1` branch.
2) Do a release of the `v1` branch
3) Bump the version in swagger-parser master to use this new version `swagger-parser-v2-version`

This is very blocking. You will be able to close this one: https://github.com/swagger-api/swagger-parser/issues/1146 at the same time.

gracekarina:
@JFCote can you please file the PR so I can review? Thanks

JFCote:
@gracekarina Well we are in it right now ;) It is just not applied to the right branch. It needs to be applied to the `v1` branch :)

JFCote:
@jmini @gracekarina I will see what I can do.

JFCote:
@gracekarina I've done a pull request here for the `v1` branch: https://github.com/swagger-api/swagger-parser/pull/1154

Issue #995, opened 2019-01-22
Parser is not resolving enum in reference discriminator


gracekarina:
related to #927 
Model Product discriminator field productType is referenced to $ref: '#/definitions/ProductType'. which contains the enum which will be extending this model.

In this case extended model does not come in resolved YAML.

````
swagger: '2.0'
info:
  version: '1.0'
  title: 'Product Operation'
  description: 'Product Operation'


basePath: /commerce

consumes:
  - application/json
produces:
  - application/json

paths:

  /v1/product/{productId}:
    get:
      summary: Get the product
      description: Get the product
      parameters:
        - name: productId
          in: path
          required: true
          type: string
      responses:
        200:
          description: OK
          schema:
            $ref: 'commonDefinitionDomain.yaml#/definitions/Product'
        400:
          $ref: 'commonDefinitionDomain.yaml#/responses/400-BadRequest'
````
````
info:
  description: "This is a common Domain"
  version: "1.0"
  title: Common Domain

responses:

  400-BadRequest:
    x-dox-Since: v10.2
    description: Bad Request
    schema:
      $ref: "#/definitions/ErrorResponse"

definitions:
  ErrorResponse:
    x-dox-Since: v10.2
    type: object
    description: |
      Response containing descriptive error text, error code
    required:
    - code
    - message
    properties:
        code:
          description: |
            The code associated with the error
          readOnly: true
          type: integer
        message:
          description: |
            The message associated with the error
          readOnly: true
          type: string

  ChannelType:
    type: string
    enum:
      - selfService
      - retail
      - callCenter
      
  Product:
    x-dox-Since: '10.2'
    description: |
      The respresentation of the product
    type: object
    discriminator: productType
    required: 
      - productType
    x-dox-discriminator-name: NAME
    x-dox-discriminator-type:
      - MobileProduct
      - InternetProduct
      - FixedVoiceProduct
    properties:
      id:
        type: string
      status:
        type: string
      name:
        type: string
      productType:
        $ref: '#/definitions/ProductType'
    
  ProductType:
    x-dox-Since: '10.2'
    description: |
      The different type of products
    enum:
      - MobileProduct
      - FixedVoiceProduct
      - InternetProduct

  MobileProduct:
    x-dox-Since: 10.2
    description: |
      The instance of the mobile product being added.
    type: object
    allOf:
      - $ref: "#/definitions/Product"
      - type: object
        properties:
          sim:
            type: string
          phoneNumber:
            type: string

[...]

PR #1004, opened 2019-01-25
refs #995 - fixing discriminator with remote enum reference

gracekarina:
refs #995 - fixing discriminator with remote enum reference

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
@@ -208,8 +208,17 @@ public final class ExternalRefProcessor {
                             processRefProperty(new RefProperty(RefType.DEFINITION.getInternalPrefix()+name), file);
                         }
                     }
-
-
+                }else if (prop.getValue() instanceof RefProperty) {
+                    String ref = ((RefProperty) prop.getValue()).getSimpleRef();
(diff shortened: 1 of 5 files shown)


modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
```diff
@@ -208,8 +208,17 @@ private void processDiscriminator(String discriminator, Map<String, Property> pr
                             processRefProperty(new RefProperty(RefType.DEFINITION.getInternalPrefix()+name), file);
                         }
                     }
-
-
+                }else if (prop.getValue() instanceof RefProperty) {
```
frantuma: fix formatting add space after bracket

modules/swagger-parser/src/main/java/io/swagger/parser/processors/ExternalRefProcessor.java
```diff
@@ -208,8 +208,17 @@ private void processDiscriminator(String discriminator, Map<String, Property> pr
...
+                    String ref = ((RefProperty) prop.getValue()).getSimpleRef();
+                    Object resolved = swagger.getDefinitions().get(ref);
```
frantuma: does this mean that the ref must have been already resolved? are we sure it will be so?

frantuma:
please build

frantuma:
please build

Issue #869, opened 2018-10-10
Update documentation explaining flatten and other options of Parser


JehandadK:
@gracekarina Not sure where should we do it? Wiki or README.md ?

frantuma:
Hey @JehandadK, I'd say README.md

PR #1011, opened 2019-02-02
Adding explanation of parser Options to readme

gracekarina:
Explaining parser options refs #869


bcoughlan:
It would be good to understand the use cases for parser options, e.g. why would somebody want to use the `flatten` property?

frantuma:
included and replaced by #1510

PR #1013, opened 2019-02-05
Elevate log level to WARNING when a conflicting Example is found

### LINKED ISSUE #1007: Elevate log level to WARNING when a conflicting Example is found
When `Example` models with same names are present in different refs, the last one that's processed is used.

The [ExternalRefProcessor](https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java) class does log a debug message ([line 76](https://github.com/swagger-api/swagger-parser/blob/d0aa5a4f93aba5097609b5d7ca586fe3cd5e6918/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java#L76)). However, it would be more useful if the message is elevated to `WARNING` level.

timemanx:
Fixes #1007


gracekarina:
@timemanx thanks for this PR, I'll have to check first how will this affect the projects that depend on parser, this should not be a blocker, and I'm not so sure this implementation won't. @frantuma

timemanx:
@gracekarina Oh, so you mean changing the log level might become a blocker for dependent projects?

Anything I can help with?

PR #1033, opened 2019-03-03
Refine creation of Schema to fix #1032

### LINKED ISSUE #1032: OpenAPI v2 converter - parse IntegerSchema instead of generic Schema
The issue reported [here](https://github.com/swagger-api/swagger-codegen/issues/8874) seems to happen due to the fact that the OpenAPI v2 converter only parses Schema, but never IntegerSchema (see [here](https://github.com/swagger-api/swagger-parser/blob/8be209a5394616ab0d074053d38e6098c687b03e/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java#L1063)). This causes generators (e.g. [here](https://github.com/swagger-api/swagger-codegen-generators/blob/151d8872efbeb71f3742e817005089a9f4054f99/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java#L1037)) to generate integer instead of long.
This original issue however does not occur with openapi 3.0.0, because the corresponding parser properly parses IntegerSchema (e.g. [here](https://github.com/swagger-api/swagger-parser/blob/8be209a5394616ab0d074053d38e6098c687b03e/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java#L2044)).
Creating/parsing Schema in SwaggerConverter the same way as in OpenAPIDeserializer (through SchemaTypeUtil) makes the original issue go away.

barney2k7:
fixes #1032 (and https://github.com/swagger-api/swagger-codegen/issues/8874)

### DIFF
--- a/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
+++ b/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
@@ -1060,7 +1060,7 @@ public class SwaggerConverter implements SwaggerParserExtension {
 
                 schema = a;
             } else {
-                schema = new Schema();
+                schema = SchemaTypeUtil.createSchema(sp.getType(), sp.getFormat());
                 schema.setType(sp.getType());
                 schema.setFormat(sp.getFormat());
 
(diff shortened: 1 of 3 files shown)


gracekarina:
@barney2k7 Thanks

PR #1042, opened 2019-03-13
#1040 Fixed duplication of external reference path

### LINKED ISSUE #1040: Duplicated relative path for model property
Location to external schema file is duplicated when external model references another model in same file. The same scenario works in previous version `2.0.8` of `swagger-parser-v3`.

Here is example:
```
.
├── lib
│   └── lib.yaml
└── api.yaml
```

api.yaml:
```yaml
openapi: 3.0.0
info:
  title: Test
  version: 1.0.0

paths:
  /value:
    get:
      operationId: getValues
      responses:
        200:
          description: Successful response
          content:
            application/json:
              schema:
                $ref: 'lib/lib.yaml#/components/schemas/Value'
```

lib/lib.yaml:
```yaml
components:
  schemas:
    ValueId:
      type: object
      properties:
        id:
          type: integer
          format: int64
    Value:
      type: object
      properties:
        id:
          $ref: '#/components/schemas/ValueId'
```

When parsed `$ref` of `id` property of `Value` model is set to:
```
/lib/lib/lib.yaml#/components/schemas/ValueId
```

Which causes:
```
Unable to load RELATIVE ref: ./lib/lib/lib.yaml path:
```
exception.

jakubmalek:
* Fix #1040 issue - duplicated reference path

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -20,6 +20,8 @@ import io.swagger.v3.oas.models.security.SecurityScheme;
 import io.swagger.v3.parser.ResolverCache;
 import io.swagger.v3.parser.models.RefFormat;
 import io.swagger.v3.parser.models.RefType;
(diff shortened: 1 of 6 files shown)


gracekarina:
Thanks @jakubmalek

Issue #1053, opened 2019-03-25
Release 2.0.10 uses SNAPSHOT dependencies


slinkydeveloper:
Hi,
Is it ok that release 2.0.10 uses snapshots from transitive dep swagger-parser 1.0.43?

```
[INFO] +- io.swagger.parser.v3:swagger-parser:jar:2.0.10:compile
[INFO] |  +- io.swagger.parser.v3:swagger-parser-v2-converter:jar:2.0.10:compile
[INFO] |  |  +- io.swagger:swagger-parser:jar:1.0.43:compile
[INFO] |  |  |  \- io.swagger:swagger-core:jar:1.5.23-SNAPSHOT:compile
[INFO] |  |  |     \- io.swagger:swagger-models:jar:1.5.23-SNAPSHOT:compile
[INFO] |  |  |        \- io.swagger:swagger-annotations:jar:1.5.23-SNAPSHOT:compile
[INFO] |  |  +- io.swagger:swagger-compat-spec-parser:jar:1.0.43:compile
[INFO] |  |  |  +- com.github.java-json-tools:json-schema-validator:jar:2.2.8:compile
[INFO] |  |  |  |  +- com.github.java-json-tools:json-schema-core:jar:1.2.8:compile
[INFO] |  |  |  |  |  +- org.mozilla:rhino:jar:1.7R4:compile
[INFO] |  |  |  |  |  \- com.github.fge:uri-template:jar:0.9:compile
[INFO] |  |  |  |  +- javax.mail:mailapi:jar:1.4.3:compile
[INFO] |  |  |  |  |  \- javax.activation:activation:jar:1.1:compile
[INFO] |  |  |  |  +- joda-time:joda-time:jar:2.9.7:compile
[INFO] |  |  |  |  +- com.googlecode.libphonenumber:libphonenumber:jar:8.0.0:compile
[INFO] |  |  |  |  \- net.sf.jopt-simple:jopt-simple:jar:5.0.3:compile
[INFO] |  |  |  +- com.github.fge:json-patch:jar:1.6:compile
[INFO] |  |  |  |  \- com.github.fge:jackson-coreutils:jar:1.6:compile
[INFO] |  |  |  |     \- com.github.fge:msg-simple:jar:1.1:compile
[INFO] |  |  |  |        \- com.github.fge:btf:jar:1.2:compile
[INFO] |  |  |  \- org.apache.httpcomponents:httpclient:jar:4.5.2:compile
[...]

PR #1056, opened 2019-03-26
Add maven-enforcer-plugin to prevent SNAPSHOT usage in releases

jmini:
To prevent having SNAPSHOT versions in the dependency tree in released versions, I propose to add the `maven-enforcer-plugin` to the build with the `requireReleaseDeps` activated for release build.

Because this project is using SNAPSHOT versions during development, I think that using `onlyWhenRelease = true` is the appropriate setting for this project.

Documentation: https://maven.apache.org/enforcer/enforcer-rules/requireReleaseDeps.html

Related:
* https://github.com/swagger-api/swagger-parser/issues/1053
* https://github.com/swagger-api/swagger-parser/pull/829#issuecomment-418793197

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -178,6 +178,28 @@
                     </rules>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


jmini:
Same change for `v1` branch: https://github.com/swagger-api/swagger-parser/pull/1057

Issue #1063, opened 2019-03-27
Parameter validation too strict


michaelom:
#960 introduced a bug:

This appears to be a valid definition:
```yaml
  /customers/{id}/legal-person/addresses/{addressId}:
    parameters:
      - $ref: "#/components/parameters/id"
      - $ref: "#/components/parameters/addressId"

```

But results in the following `messages` after Parsing:

```
attribute paths.'/customers/{id}/legal-person/addresses/{addressId}'.There are duplicate parameter values
```

Code:
```
 val parseResult = OpenAPIV3Parser().readWithInfo("http://localhost:8080/swagger/openapi.yaml", emptyList())

            val errors = parseResult.messages.joinToString(separator = System.getProperty("line.separator"))
            print(errors)

```

cvgaviao:
Hello @jakubmalek, I did validate your patch and confirmed that it fixed the validation messages that I was facing with multiple parameter references at #1066. 
But I'm wondering... 
Why I'm getting a parameter with null values (including its name) when I do:
```
PathItem path = api.getPaths().get("/anPath/{anParameter}");
Parameter anParam = path.getParameters().get(0);
assertThat(anParam.getName()).isEqualTo("anParameter");
```  
those operation parameters wouldn't be resolved at parsing stage ?

gracekarina:
closed by #1065 and #1080

PR #1065, opened 2019-03-29
Fixed referenced parameters validation

jakubmalek:
Fix for issue #1063.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -1367,6 +1367,8 @@ public class OpenAPIDeserializer {
             return null;
         }
         Map<String, Parameter> parameters = new LinkedHashMap<>();
+        Set<String> filter = new HashSet<>();
+        Parameter parameter=null;
 
         Set<String> parameterKeys = getKeys(obj);
         for(String parameterName : parameterKeys) {
@@ -1379,12 +1381,13 @@ public class OpenAPIDeserializer {
             if (parameterValue.getNodeType().equals(JsonNodeType.OBJECT)) {
                 ObjectNode parameterObj = (ObjectNode) parameterValue;
                 if(parameterObj != null) {
-                    Parameter parameter = getParameter(parameterObj, String.format("%s.%s", location, parameterName), result);
+                     parameter = getParameter(parameterObj, String.format("%s.%s", location, parameterName), result);
                     if (parameter != null) {
                         parameters.put(parameterName, parameter);
                     }
                 }
             }
+
         }
         return parameters;
     }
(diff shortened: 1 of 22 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -1408,13 +1412,25 @@ else if(!v.isValueNode()) {
         }
         Set<String> filter = new HashSet<>();
 
-        for(Parameter param:parameters) {
+        parameters.stream().map(this::getParameterDefinition).forEach(param -> {
             if(!filter.add(param.getName()+"#"+param.getIn())) {
```
cvgaviao: @gracekarina , this is not working when the parameter is coming from another file. param.getName() will always return null in this case, so it won't pass the next filter.

gracekarina: Hi please, send us a spec where this will fail so we can test and addressed the issue.

cvgaviao: File one:
```
openapi: 3.0.0
info:
  version: 1.0.0
  title: an API
  description: An API for reproduce a *There are duplicate parameter values*.

paths:
  /anPath:
    get:
      parameters:
        - $ref: './references.oas3.yaml#/components/parameters/ParamQ_One'
        - $ref: './references.oas3.yaml#/components/parameters/ParamQ_Two'
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
```
file references.oas3.yaml:
```
components:
  parameters:
    ParamQ_One:
      in: query
      name: customer-id
      required: true
      schema:
        format: int32
        type: integer
    ParamQ_Two:
      in: query
      name: unit-id
      schema:
        format: int32
        type: integer
```

gracekarina: well, to resolve parameters it should be done after deserializing in the resolver part. I'm looking at it, but this is a different issue.

cvgaviao: @gracekarina, 
well, could be, but I don't think it is a different issue...
we'll still get 'duplicated values' because a parameter is only unique when we have its name + its type. if we don't have its type then the problem of duplicate will arise and will preventing any generator to work.

gracekarina: What could be done is validate in deserializing time, if the ref in the parameter is internal, or remote so we would avoid the message warning.

cvgaviao: Agreed, and I don't see any option other than to resolve that $ref at this deserializing time... :)

gracekarina: merged PR https://github.com/swagger-api/swagger-parser/pull/1080

cvgaviao: @gracekarina, well done, girl !

PR #1076, opened 2019-04-14
fixed set types to schemas 

gracekarina:
fix for issue #1039º


jmini:
I think this change in `OpenAPIDeserializer` creates this new issue: #1090.

I propose PR #1092 to fix it.

Issue #1087, opened 2019-04-26
v1 parser not resolving common parameters that use relative references


cmuchinsky:
Using the v1 parser, top level parameters that use relative references are not being resolved. Take for example parsing the following swagger fragment:
```
...
parameters:
  param2:
    $ref: "./parameters/params.yaml#/param2"
...
```
Actual results:
```
...
parameters:
  param2:
    $ref: "./parameters/params.yaml#/param2"
...
```
Expected results:
```
...
parameters:
  param2:
    name: "param2"
    in: "header"
    required: false
    type: "string"
...
```

gracekarina:
closed by #1088

cmuchinsky:
@gracekarina any thoughts on when a v1 release with this fix might be coming?

PR #1088, opened 2019-04-26
Ensure top level parameter relative references are resolved

cmuchinsky:
swagger-api/swagger-parser#1087


gracekarina:
thanks @cmuchinsky

Issue #1090, opened 2019-05-03
2.0.12 regression: integer type replaced with number when enum values are present


jmini:
When an object properties has enum value defined:

```yaml
openapi: 3.0.1
info:
  title: Object with enum test
  version: '1.0'
servers:
  - url: 'http://localhost:9999/'
paths:
  /ping:
    get:
      operationId: pingGet
      responses:
        '201':
          description: OK
components:
  schemas:
    SomeObj:
      type: object
      properties:
        iprop:
          type: integer
          format: int32
          enum:
            - 1
            - 2
        lprop:
          type: integer
          format: int64
          enum:
            - 1
            - 2
        nprop:
          type: number
          format: double
          enum:
            - 1.1
            - 1.2
```

With this code:
```java
OpenAPIParser openApiParser = new OpenAPIParser();
ParseOptions options = new ParseOptions();
options.setResolve(true);
options.setFlatten(true);

SwaggerParseResult parseResult = openApiParser.readLocation(inputSpec, null, options);
OpenAPI openAPI = parseResult.getOpenAPI();
Schema someObj = openAPI.getComponents().getSchemas().get("SomeObj");
Map<String, Schema> properties = someObj.getProperties();

Schema iprop = properties.get("iprop");
System.out.println("iprop: " + iprop.getType());

Schema lprop = properties.get("lprop");
System.out.println("lprop: " + lprop.getType());

Schema nprop = properties.get("nprop");
System.out.println("nprop: " + nprop.getType());
```
---
With version `2.0.11` you get the expected result:
```
iprop: integer
lprop: integer
nprop: number
```

---
With version `2.0.12` it does not work:
```
iprop: number
lprop: number
nprop: number
```

=> The type is not the correct one.

gracekarina:
thanks! 
Closed by #1092

PR #1092, opened 2019-05-04
Fix for issue #1090

jmini:
Extract the `StringUtils.isBlank(schema.getType())` condition from the if introduced with #1076.

Fix for issue #1090.


gracekarina:
thanks @jmini.

Issue #997, opened 2019-01-24
Properties order lost while converting from v2 to v3


fmoussallam:
While both v2 and v3 models store properties in LinkedHashMaps, io.swagger.v3.parser.converter.SwaggerConverter#convert(Property schema) uses a HashMap as temporary storage, losing the properties order.

This is for version 2.0.6 but I can see issue still there on master: https://github.com/swagger-api/swagger-parser/blob/a3d3a4207ded6414573ea4d2986024f931f364f9/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java#L924

Thanks

jmini:
I am not maintainer of this project, but I approve such a change.

A lot of `HashMap` in Swagger-core were replaced by `LinkedHashMap`.

PR #1093, opened 2019-05-05
Properties order lost while converting from v2 to v3 #997

fmoussallam:
**What is the problem:** In our unit tests,  properties need to appear in a specific order. We are having issues when these tests are run with different JDKs because HashMap ordering is not guaranteed.

**What is the solution:** Replace HashMap with LinkedHashMap in SwaggerConverter.java.

Please note that a unit test for this PR would be subject to the same JDK issue mentioned here, so no unit test is provided.


gracekarina:
thanks!

Issue #1086, opened 2019-04-25
Number type MultipleOf fields are not being parsed and are always null


Leo11221:
The "multipleOf" fields for number types are not being parsed, they always seem to be null have attached a Scala unit test to replicate:

```
import io.swagger.v3.parser.OpenAPIV3Parser
import io.swagger.v3.parser.core.models.ParseOptions


class MultipleOfTest {

  "multiple of" should "be not null" in
  {
    val openApiDef =
      """openapi: 3.0.0
        |info:
        |  version: 0.0.0
        |  title: "test"
        |components:
        |  schemas:
        |    AssessCandidate:
        |      type: object
        |      properties:
        |        candidate:
        |          type: object
        |          properties:
        |            id:
        |              type: string
        |          required:
        |          - id
        |        test_results:
        |          type: object
        |          properties:
        |            id:
        |              type: string
        |            status:
        |              type: string
        |            score:
        |              type: number
        |              multipleOf: 1
        |            date:
        |              type: string
        |              format: date
        |            comment:
        |              type: string
        |            result_url:
        |              type: string
        |      required:
        |      - candidate
        |      - test_results
        |""".stripMargin
    
    val multipleOf = new OpenAPIV3Parser().readContents(openApiDef, null, new ParseOptions).getOpenAPI.getComponents.getSchemas.get("AssessCandidate").getProperties.get("score").getMultipleOf
    assert(multipleOf != null)
    }
  }
```

ymohdriz:
Hi @Leo11221 

We tried to reproduce the issue, it's not reproducible and newly added tests are passing.
It seems that way you tried to read the 'score' property is incorrect. Please have a look at the TC #1095  



Thanks,
Mohammed

gracekarina:
Hi, @Leo11221 is the PR #1095 helping to clarify this issue? if this is still an issue, can you please provide a PR so we can reproduce and fix? thanks

PR #1095, opened 2019-05-09
Test case for issue 1086

ymohdriz:
Hi team,

This test indicates that the issue mentioned in #1086 is not reproduced.

Thanks,
Mohammed


gracekarina:
thanks @ymohdriz.

PR #1098, opened 2019-05-14
xml information in array models

Britz:
The xml information was ignored while creation of array models... So I added them. 

This is e.g. relevant when using openapi-generator (3.3.4 in my case) to generate custom java classes from swagger2.0 descriptions with correct xml annotations.


gracekarina:
@Britz thanks!

Issue #1113, opened 2019-06-03
OpenAPI v2 converter - incorrect server URL when basePath begins with a '/'


alexweav:
I have a v2 document with a `basePath` of `/nitest`, no `host`, and no `schemes`. The v2 converter prepends two forward slashes to it, returning a Server URL of `///nitest`, which is incorrect - i would expect it to be unchanged.

This prepend occurs in [SwaggerConverter.java](https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java), line 424.

With a bit of guidance, I can send a PR to fix it. It seems to me that there needs to be an extra case in this method for basePaths which begin with a `/` when no schemes are available.

If this is the correct approach, or if we should take a different route, let me know - I can post a PR either way.

alexweav:
I added a test for this, plus a fix, in separate PRs.

gracekarina:
Hey, sorry I will be checking this pr in the next days.

gracekarina:
Closed by #1115

alexweav:
Thank you!

PR #1114, opened 2019-06-03
Add test for Issue 1113

alexweav:
Adds a test case for #1113 where the OpenAPI v2 Converter adds two leading slashes to the baseUrl when the baseUrl begins with a slash and there are no schemes present. Test currently fails because the resulting URL is `///test`.


gracekarina:
hi! @alexweav is this PR still needed for merged, or was it included on #1115 ?

alexweav:
@gracekarina this was included in #1115. Thanks very much!

PR #1115, opened 2019-06-03
Avoid prepending forward slashes to URLs when the URL already starts with a forward slash

alexweav:
Fixes issue #1113. We no longer prepend `//` to the `basePath` when the `basePath` already begins with a `/`. Fixes (and contains) the failing test added in #1114.


gracekarina:
@alexweav thanks for this PR!.

PR #1134, opened 2019-06-26
issue-1133 - Parser does not resolve $ref pointing to relative locations on classpath

scottr-ad:
PathUtils - improve url lookup to find classpath resources not in swagger-parser jars. 
ResolverCache, RefUtils - add lookup by classpath when file path not resolvable.


scottr-ad:
@gracekarina Could you or one of the other project contributors please review this PR and merge it or provide feedback?

PR #1162, opened 2019-08-01
issue-1161 - ResolverFully does not resolve $ref in ComposedSchema with properties

scottr-ad:


scottr-ad:
@gracekarina Could you or one of the regular committers please review this PR when you get a chance?

PR #1165, opened 2019-08-01
Using specialized schema subclasses

### LINKED ISSUE #1164: Swagger v2 converter does not use specialized Schema subclasses
The swagger converter just does `new Schema`, then `setType` and `setFormat` manually afterwards.

This prevents consumers from being able to `isInstanceOf[StringSchema]`, requiring `isInstanceOf[StringSchema] && schema.getType() == "string"` everywhere, defeating the purpose of having those specialized schemas in the first place.

blast-hardcheese:
Resolves #1164 

Motivation:
`PrimitiveType` contains a full mapping (and constructors!) for built-in tracked schema subtypes. By using this, the Schemas generated by the 2to3 converter will have a similar shape to the native 3.x schemas, permitting more unified structure handling for clients (permitting relying on `isInstanceOf` exclusively, instead of `isInstanceOf` as well as `getType() == "xyz"`)

Delegating to `fromTypeAndFormat` to get a particular instance, then `createProperty()` to instantiate the particular schema resolves the issue I was running into.

If we can't actually look up the mapping, fall back to the original behaviour.

### DIFF
--- a/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
+++ b/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
@@ -32,6 +32,7 @@ import io.swagger.parser.SwaggerParser;
 import io.swagger.parser.SwaggerResolver;
 import io.swagger.parser.util.SwaggerDeserializationResult;
 import io.swagger.v3.core.util.Json;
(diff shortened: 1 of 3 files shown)


blast-hardcheese:
I force pushed because I found a fallback case where the base type is recognized by `PrimitiveType`, but the format was not. I handled this by calling with `null` as the format, then `setFormat` on the resulting schema (if it exists) otherwise continuing down the chain to `new Schema()` with the previous behaviour.

PR #1229, opened 2019-10-17
added failing test case for issue-1228

r-sreesaran:


r-sreesaran:
This pr uses a value for minimum parameter as string  rather than double

PR #1245, opened 2019-11-05
Issue with resolving a external $ref in a relative subfolder.

joeljons:
This is not a suggested fix, just a PR to reproduce a problem I ran into with a minimal example.

I have a ref like `$ref: './subfolder/domain.yaml#/components/responses/ErrorResponse`
When resolving I get the following error `Unable to load RELATIVE ref: ./subfolder/subfolder/domain.yaml`
Note the double ./subfolder/subfolder! Where did it come from?


gracekarina:
addressed by #1247

joeljons:
Thanks!

PR #1251, opened 2019-11-13
Remove unused slf4j dependencies from pom

mennopruijssers:
The slf4j-ext and slf4j-api dependencies is actually not used. This PR removes it from the pom files.
Discovered it because `slf4j-ext` is banned from one of the projects I'm working on.


gaiazov:
Would love to see this merged!

Issue #1273, opened 2019-12-19
OpenAPIDeserializer produces message with location ...responses.responses...


brhaible:
The OpenAPIDeserializer is generally good at indicating where an error was encountered. However, in the case below, it produces a message
```
attribute paths.'/myPath'(get).responses.responses.bar is not of type `object`
```
It should instead read
```
attribute paths.'/myPath'(get).responses.bar is not of type `object`
```

How to reproduce:
1. Store this file as input.yaml.

[input.yaml.gz](https://github.com/swagger-api/swagger-parser/files/3985005/input.yaml.gz)

2. Store this file as ParseOpenAPI3File.java.

[ParseOpenAPI3File.java.gz](https://github.com/swagger-api/swagger-parser/files/3985010/ParseOpenAPI3File.java.gz)

Compile it with a classpath that contains
slf4j-api-1.5.8.jar
slf4j-jdk14-1.5.8.jar
jackson-core-2.9.9.jar
jackson-dataformat-yaml-2.9.9.jar
jackson-databind-2.9.9.jar
jackson-annotations-2.9.9.jar
commons-io-1.4.jar
commons-lang3-3.1.jar
swagger-models-2.0.8.jar
swagger-parser-v3-2.0.12.jar
swagger-core-2.0.8.jar
swagger-parser-core-2.0.12.jar
snakeyaml.jar

3. Run it with one argument, namely the absolute file name from step 1.

brhaible:
The following patch fixes it:

[issue1273-fix.diff.gz](https://github.com/swagger-api/swagger-parser/files/3985030/issue1273-fix.diff.gz)

gracekarina:
closed by #1294

PR #1294, opened 2020-01-08
Fix duplicate '.responses' in location string.

brhaible:
Issue: https://github.com/swagger-api/swagger-parser/issues/1273


gracekarina:
thanks! @brhaible

Issue #1274, opened 2019-12-19
OpenAPIDeserializer produces message with location that omits the mediaType


brhaible:
The OpenAPIDeserializer is generally good at indicating where an error was encountered. However, in the case below, it produces a message
```
attribute paths.'/myPath'(get).responses.200.content.bar is unexpected
```
It should instead read
```
attribute paths.'/myPath'(get).responses.200.content.'application/json'.bar is unexpected
```

How to reproduce:

1. Store this file as input.yaml.

[input.yaml.gz](https://github.com/swagger-api/swagger-parser/files/3985051/input.yaml.gz)

2. Store this file as ParseOpenAPI3File.java.

[ParseOpenAPI3File.java.gz](https://github.com/swagger-api/swagger-parser/files/3985053/ParseOpenAPI3File.java.gz)

Compile it with a classpath that contains
slf4j-api-1.5.8.jar
slf4j-jdk14-1.5.8.jar
jackson-core-2.9.9.jar
jackson-dataformat-yaml-2.9.9.jar
jackson-databind-2.9.9.jar
jackson-annotations-2.9.9.jar
commons-io-1.4.jar
commons-lang3-3.1.jar
swagger-models-2.0.8.jar
swagger-parser-v3-2.0.12.jar
swagger-core-2.0.8.jar
swagger-parser-core-2.0.12.jar
snakeyaml.jar

3. Run it with one argument, namely the absolute file name from step 1.

brhaible:
The following patch fixes it:

[issue1274-fix.diff.gz](https://github.com/swagger-api/swagger-parser/files/3985069/issue1274-fix.diff.gz)

webron:
@brhaible Thanks for all your reports recently, we greatly appreciate it. For the ones you have a fix for, would you mind submitting a PR?

brhaible:
_> For the ones you have a fix for, would you mind submitting a PR?_
Done.

webron:
Thanks! Much appreciated.

gracekarina:
closed by #1295

PR #1295, opened 2020-01-08
Inside a content, include the mediaType in the location string.

brhaible:
Issue: https://github.com/swagger-api/swagger-parser/issues/1274


gracekarina:
thanks!! @brhaible

Issue #1305, opened 2020-01-22
security not parsing "none" (branch v1)


wsalembi:
Same as #492 bug already resolved in v3

empty security object is not getting parsed

"security": [
  {},
  {"oauth": […]},
  {"token-1": []},
  {"token-2": []}
]
parsed has only 3 elements

PR #1306, opened 2020-01-22
#1305 fixing parsing of empty securityRequirement

wsalembi:
Because in the v1 library operation.getSecurity() doesn't return a list of SecurityRequirements cfr the top-level security field in the swagger file, I opted for adding an empty Map to the list to represent the none security requirement.

We need swagger-parser to retain the unsecured security requirement on an operation level. Now it is stripped


gracekarina:
thanks @wsalembi

Issue #1249, opened 2019-11-12
Swagger 2 validation allows path params to be non-required


gexclaude:
As the spec states, path params are always required:
https://swagger.io/docs/specification/2-0/describing-parameters/

We encountered, that this validation is not performed in swagger-parser version 1.0.47.

A small example defining a path with two path params - one is required=true and one is required=false. This appears valid when reading with `new SwaggerParser().readWithInfo(content);`

```
{
  "swagger": "2.0",
  "info": {
	"version": "1.0.0",
	"title": "Example API"
  },
  "host": "example.com",
  "basePath": "/",
  "paths": {
	"/api/admin/{account}/users/{user}": {
	  "post": {
		"operationId": "addUser",
		"parameters": [
		  {
			"name": "account",
			"in": "path",
			"required": true,
			"type": "string"
		  },
		  {
			"name": "user",
			"in": "path",
			"required": false,
			"type": "string"
		  }
		],
		"responses": {
		  "200": {
			"description": "OK"
		  },
		}
	  }
	},
  }
}
```

However editor complains, which is correct in my opinion:
![image](https://user-images.githubusercontent.com/416252/68694805-20a42a00-057a-11ea-9f7c-179bdc02c89c.png)

gexclaude:
It may relate to https://github.com/swagger-api/swagger-parser/issues/1111

gexclaude:
It appears that the parser does not perform validation against the JSON schema. Is this intentional?

GORA-SAG:
Hi @gexclaude ,
As you can see the issue has been fixed. The fix will be available in 1.0.50. So, Kindly mark it as closed.

Regards,
Gokul Raj S

gexclaude:
Thank you a lot for your efforts!

PR #1308, opened 2020-01-31
Fix for issue #1249

GORA-SAG:
Swagger 2 validation allows path params to be non-required


GORA-SAG:
Hi @ymohdriz ,
Sure, i will do.

Regards,
GORA

PR #1312, opened 2020-02-07
Fix for issue #1249

GORA-SAG:
Previously we didn't take the reference parameter into account. Now added warning for reference parameter also.


gracekarina:
thanks @GORA-SAG

PR #1271, opened 2019-12-09
enable STRICT_DUPLICATE_DETECTION in Jackson parser

sullis:

PR #1313, opened 2020-02-07
enable STRICT_DUPLICATE_DETECTION in Jackson parser for Yaml

jmini:
This PR is doing the same change as @sullis did in #1271 for JSON but for the YAML format.

Without the `STRICT_DUPLICATE_DETECTION`, when there are duplicates keys Jackson is silently ignoring the first one. This results in unpredictable behavior.

### Example

Following Yaml is invalid because of the 2 `"200"` keys in the response map:
```yaml
openapi: 3.0.1
info:
  title: title
  description: This is a sample server
  license:
    name: Apache-2.0
    url: http://www.apache.org/licenses/LICENSE-2.0.html
  version: 1.0.0
servers:
- url: https://api.absolute.org/v2
  description: An absolute path
paths:
  /whatever:
    get:
      summary: Some operation
      description: Some operation
      operationId: doWhatever
      responses:
        "200":
          description: OK
        "200":
          description: duplicate HTTP status code
```

As comparison the Swagger-Editor Javascript parser is doing a better job and displays an error:

image

### Other example

```yaml
openapi: 3.0.1
info:
  title: test with duplicate keys
  version: '1.0'
paths:
  /ping:
    get:
      operationId: pingOperation
      x-foo: bar
      x-foo: baz
      responses:
        '201':
          description: OK
```

With this example `getVendorExtensions()` is returning only one of the two `x-foo` keys.

After the change of this PR, an error is thrown by parsing (which is good because the Yaml is invalid in this case).

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ObjectMapperFactory.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ObjectMapperFactory.java
@@ -15,7 +15,7 @@ public class ObjectMapperFactory {
     }
 
     protected static ObjectMapper createJson(boolean includePathDeserializer, boolean includeResponseDeserializer) {
-        return create(null, includePathDeserializer, includeResponseDeserializer);
+        return create(createJsonFactory(), includePathDeserializer, includeResponseDeserializer);
     }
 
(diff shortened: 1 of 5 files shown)


sullis:
nice work!  @jmini

Issue #1311, opened 2020-02-06
Deserialization of OpenAPI v3 Schema fails


jpommerening:
Hi everyone!

I think I found a very small issue in the OpenAPIDeserializer; the JSON schema parser more specifically.

**tl;dr: The parser throws a NullPointerException when it encounters a JSON schema that uses `patternProperties`.**

First, let me try to explain what happened: I'm describing an endpoint which returns an OpenAPI v3 document ("swagger.yaml") as a response.
To describe the response format, I'm `$ref`ering to a local copy of the [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/tree/master/schemas/v3.0).

When using swagger-parser to parse it as a schema, I get a `NullPointerException` in `io.swagger.v3.parser.util.OpenAPIDeserializer.getSchema(OpenAPIDeserializer.java:2418)` (which is [here](https://github.com/swagger-api/swagger-parser/blob/0439c63c97843f3a773a6802a693399fc4d7f387/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java#L2423)).

Turns out, the Parser is trying to access `result.extra` when `result` is `null`.
I'm not *exactly* sure what that signifies, but I discovered that the `patternProperties` key was causing the error.
If I understand that part of the method correctly, I'd assume that `patternProperties` should be listed in [`SCHEMA_KEYS`](https://github.com/swagger-api/swagger-parser/blob/0439c63c97843f3a773a6802a693399fc4d7f387/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java#L89).

The following should file should trigger the NPE:

```yaml
openapi: 3.0.0
info:
  title: Deserialization of OpenAPI v3 Schema fails
  version: 0.1.0
paths:
  /:
    get:
      description: When the response schema uses patternProperties
      responses:
        500:
          description: NPE
          content:
            'application/json':
              schema:
                $ref: 'https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/schemas/v3.0/schema.yaml'
```

*Edit:* updated example yaml. Apparently, it only happens when using `$ref`.

webron:
Technically, that makes is an invalid OpenAPI definition as there's no support for full JSON Schema and some keywords are not supported (this will change in OAS3.1).

jpommerening:
[...]

PR #1314, opened 2020-02-07
Fix for #1311

jpommerening:
While this does not add the capabilities to store `patternProperties` in ObjectSchemas, it fixes issue #1311 by at least making sure, that schema deserialization does not throw in that particular case.


gracekarina:
closing due to close issue

Issue #1369, opened 2020-05-15
additionalProperties not set correctly if omitted or a boolean


spacether:
Per the openapi v2 and greater spec definitions, additionalProperties should be set to empty schema {} if it is omitted. It is currently being set to null.

v2.0 points us to the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md JSON schema Draft 4 definition, and OpenAPI does not tell us a default value.
Per [here](https://stackoverflow.com/questions/16459954/understanding-the-additionalproperties-keyword-in-json-schema-draft-version-4) it sounds like an omitted additionalProperties should be an empty schema.

v3.0.0 points us to the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md JSON Schema Specification Wright Draft 00 definition.
Wright Draft 00 mentions
If "additionalProperties" is absent, it may be considered present with an empty schema as a value.
v3.0.3 explicitly says that it defaults to true here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md

Specifically from https://tools.ietf.org/html/draft-wright-json-schema-validation-00#page-10:
```
If "additionalProperties" is absent, it may be considered present with an empty schema as a value.
```

When additionalProperties is omited, how about we set it to {} for:
- object type schemas
- composed schemas
So that we conform to the specs

gracekarina:
Hi @spacether both PR are failing, please fix so we can review and merge. thanks!

spacether:
Thanks for pointing that out; just now I fixed the tests in my https://github.com/swagger-api/swagger-parser/pull/1414 PR

webron:
Jumping in a little bit late, but better late than...

While what you mention for v3 is true, for v2 it's not so much.

v2 has several inconsistencies and some missing information. When it comes to JSON Schema, it was due to the fact that nobody really knew JSON Schema well when the decision was made to use it.

In v2, _despite what the specification might say_, the default for additionalProperties was assumed to be`false` and `true`/`false` value were not allowed. If you wanted to allow any additional property, you explicitly had to set it to an empty schema. As you probably already know, this has changed in v3, and have been better clarified there.

I don't think we should change the behavior for v2 - it should become more and more obsolete especially with the upcoming release of v3.1.

spacether:
@webron do you have any references that show that decision/interpretation for v2?

webron:
[...]

PR #1370, opened 2020-05-15
[1369] Unit test for AdditionalProperties keyword when converting from v2 to v3

sebastien-rosset:
This is a fix for #1369. WIP

### DIFF
--- a/modules/swagger-parser-v2-converter/src/test/java/io/swagger/parser/test/V2ConverterTest.java
+++ b/modules/swagger-parser-v2-converter/src/test/java/io/swagger/parser/test/V2ConverterTest.java
@@ -857,12 +857,23 @@ public class V2ConverterTest {
     public void testissue1369() throws Exception {
         OpenAPI oas = getConvertedOpenAPIFromJsonFile(ISSUE_1369_YAML);
         assertNotNull(oas);
-        Schema schema = (Schema) oas.getComponents().getSchemas().get("S1");
-        assertNotNull(schema.getAdditionalProperties());
-        schema = (Schema) oas.getComponents().getSchemas().get("S2");
-        assertNotNull(schema.getAdditionalProperties());
-        schema = (Schema) oas.getComponents().getSchemas().get("S3");
+        Schema schema = (Schema) oas.getComponents().getSchemas().get("Foo1");
+        assertNotNull(schema);
         assertNull(schema.getAdditionalProperties());
+
+        schema = (Schema) oas.getComponents().getSchemas().get("Foo2");
+        assertNotNull(schema);
+        assertNotNull(schema.getAdditionalProperties());
+        assertEquals(schema.getAdditionalProperties(), Boolean.TRUE);
+
+        schema = (Schema) oas.getComponents().getSchemas().get("Foo3");
+        assertNotNull(schema);
+        assertNotNull(schema.getAdditionalProperties());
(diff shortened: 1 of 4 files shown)


spacether:
@gracekarina any chance that this can make it in to the master branch?
This would really help downstream consumers of Swagger Parser.

JonathanParrilla:
please build

sebastien-rosset:
> please build

I have pulled the latest from master. As it stands, this PR has a unit test to show the problem. It does not have the actual fix. I initially spent some time looking at how to fix the problem, but in doing so it occurred to me the v2 code seems to have been the basis for v3 code, with some code being copied from v2 into a v3 folder, then modified to handle additional properties. Code maintainers probably have opinions on how this problem should be fixed.

spacether:
Heads up, I just submitted this PR: https://github.com/swagger-api/swagger-parser/pull/1414
Which adds unit tests and a fix for additionalProperties

modules/swagger-parser-v2-converter/src/test/java/io/swagger/parser/test/V2ConverterTest.java
```diff
@@ -852,6 +853,29 @@ public void testissue1261() throws Exception {
...
+        OpenAPI oas = getConvertedOpenAPIFromJsonFile(ISSUE_1369_YAML);
+        assertNotNull(oas);
+        Schema schema = (Schema) oas.getComponents().getSchemas().get("Foo1");
+        assertNotNull(schema);
+        assertNull(schema.getAdditionalProperties());
+
+        schema = (Schema) oas.getComponents().getSchemas().get("Foo2");
+        assertNotNull(schema);
+        assertNotNull(schema.getAdditionalProperties());
+        assertEquals(schema.getAdditionalProperties(), Boolean.TRUE);
```
spacether, line 867: getAdditionalProperties() value should be a schema a this step in processing for a true spec input

modules/swagger-parser-v2-converter/src/test/java/io/swagger/parser/test/V2ConverterTest.java
```diff
@@ -852,6 +853,29 @@ public void testissue1261() throws Exception {
...
+
+        schema = (Schema) oas.getComponents().getSchemas().get("Foo3");
+        assertNotNull(schema);
+        assertNotNull(schema.getAdditionalProperties());
+        assertEquals(schema.getAdditionalProperties(), Boolean.FALSE);
```
spacether, line 872: getAdditionalProperties() value should be null at this step in processing for a false spec input

modules/swagger-parser-v2-converter/src/test/java/io/swagger/parser/test/V2ConverterTest.java
spacether: My mistake, this should be Boolean.TRUE per checking results when parsing a v3 spec.
This is good

modules/swagger-parser-v2-converter/src/test/java/io/swagger/parser/test/V2ConverterTest.java
spacether: My mistake, this should be Boolean.FALSE per checking results when parsing a v3 spec.
This is good

PR #1374, opened 2020-05-15
Add extension to keep track of original swagger version

sebastien-rosset:
This is a proposed change to retain the original value of the "swagger" keyword when converting a OAS/Swagger 2.0 document to a 3.0 OAS document.


gracekarina:
Hi @sebastien-rosset can you please update the branch so we can merge? thanks!

sebastien-rosset:
> Hi @sebastien-rosset can you please update the branch so we can merge? thanks!

Done. Just a thought, how about renaming the extension to `x-original-openapi-version`? In the future, there may be a similar use case when version 3 is converted to version 4, so it makes more sense to use `openapi` instead of `swagger`

PR #1376, opened 2020-05-16
Fix for issue #1375

### LINKED ISSUE #1375: Empty and blank query parameter examples are parsed to null
Empty and blank query parameter examples are parsed to null, while it's expected them to be parsed to a valid empty or blank string respectively.

```
openapi: 3.0.2
info:
  title: Example references
  version: 1.0.0
paths:
  /foo:
    get:
      parameters:
        - name: paramName
          in: query
          example: " "
      responses:
        '200':
          description: success
```

`result.getOpenAPI().getPaths().get("/foo").getGet().getParameters().get(0).getExample()` returns `null` while I expect it to return `" "`

The same is with an empty string.

mih-kopylov:
Empty and blank query parameter examples were parsed to null.
Now they are parsed to an empty and blank string respectively.

Fixes #1375


mih-kopylov:
@gracekarina could you please take a look?

PR #1377, opened 2020-05-18
Improvement for issue #1303

### LINKED ISSUE #1303: Question: ParseOptions resolve versus resolveFully
This is a pretty simple question, I think.

Related to code starting here: https://github.com/swagger-api/swagger-parser/blob/d88d11c10252f98b052ee7ce690de334ab48adaf/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java#L66

```java
if (options.isResolve()) {
    result.setOpenAPI(resolver.resolve());
}
if (options.isResolveFully()) {
    result.setOpenAPI(resolver.resolve());
    new ResolverFully(options.isResolveCombinators()).resolveFully(result.getOpenAPI());
} else if (options.isFlatten()) {
    InlineModelResolver inlineModelResolver = new InlineModelResolver(options.isFlattenComposedSchemas());
    inlineModelResolver.setSkipMatches(options.isSkipMatches());
    inlineModelResolver.flatten(result.getOpenAPI());
}
```

Is it intentional for resolve=true and resolveFully=true to resolve the document twice, or is this missing an `else` condition plus resolution logic in the `isFlatten` block? I see the same behavior in `readLocation` and `readContents`, so I assume it's intentional but I don't understand the use case.

I'm looking at setting either resolve or resolveFully, but not both. I'm hesitant to do so because it looks like the parser may be implicitly relying on some caching done by the first resolve whenever resolveFully is invoked, but there are no checks for if resolve and resolveFully are both set.

mih-kopylov:
Refactor OpenAPIV3Parser to make it more clear

Fixes #1303

### DIFF
--- a/modules/swagger-parser-core/src/main/java/io/swagger/v3/parser/core/models/SwaggerParseResult.java
+++ b/modules/swagger-parser-core/src/main/java/io/swagger/v3/parser/core/models/SwaggerParseResult.java
@@ -1,6 +1,8 @@
 package io.swagger.v3.parser.core.models;
 
 import io.swagger.v3.oas.models.OpenAPI;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 public class SwaggerParseResult {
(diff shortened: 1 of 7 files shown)


gracekarina:
@mih-kopylov Thanks a lot! This is included in #1380 which adds some minor updates in #1380 for compatibility

Issue #1409, opened 2020-08-01
NPE when parameter style value is invalid


kerrykimbrough:
Found first in v3 2.0.19 and confirmed in the current snapshot.

What happens if you accidentally give an invalid value for the parameter style? For example:

```
...
"parameters": [
    {
    "name": "myParam",
    "style": "DERP",
 ...
```

At the surface, you'll see a parse result with single mysterious message: "null". Looking deeper, you'll see that the encounter with the bad style (line 1559) leads to an NPE at line 1564.

```java
 1558          value = getString("style", obj, false, location, result);
 1559          setStyle(value, parameter, location, obj, result);       // Bad style found here
 1560  
 1561          Boolean explode = getBoolean("explode", obj, false, location, result);
 1562          if (explode != null) {
 1563              parameter.setExplode(explode);
 1564          } else if(parameter.getStyle().equals(StyleEnum.FORM)){  // ...then NPE happens here
 1565              parameter.setExplode(Boolean.TRUE);
 1566          } else {
 1567              parameter.setExplode(Boolean.FALSE);
 1568          }
```

gracekarina:
close by https://github.com/swagger-api/swagger-parser/pull/1410

PR #1410, opened 2020-08-01
When parameter style is invalid, return a helpful message instead of throwing an NPE

kerrykimbrough:
This fixes the NPE (see issue #1409) and also alters the resulting error message to more accurately describe the problem.


gracekarina:
thanks @kerrykimbrough for this fix!

PR #1414, opened 2020-08-03
Fixes additionalProperties for v2 (swagger) specs

spacether:
Fixes https://github.com/swagger-api/swagger-parser/issues/1369
Brings additionalProperties values into compliance with the Openapi 2.0 specification
Fixes two CI tests which were failing. They were failing because their requests to get json files from a url were not mocked and the requested url resource was returning a http 302 status code without the spec content.

Here is what the code in the v1 branch has done before this PR:
- additionalProperties = true -> null
- additionalProperties = false -> null
- additionalProperties = string schema
- additionalProperties = refed schema
- additionalProperties omitted -> null

If this fix is merged, the code will now do the following:
- additionalProperties = true -> {} (any type)
- additionalProperties = false -> null
- additionalProperties = string schema
- additionalProperties = refed schema
- additionalProperties omitted -> {} (any type)

### TODO, update to:
- additionalProperties = true -> Boolean.TRUE
- additionalProperties = false -> Boolean.TRUE
- additionalProperties = string schema
- additionalProperties = refed schema
- additionalProperties omitted -> null
This will bring our results in line with the v3 spec parsing.

### Future Thoughts
In the future it would be helpful if parsing of v2 and v3 specs resulted in:
- additionalProperties = true -> {} (any type)
- additionalProperties = false -> null
- additionalProperties = string schema
- additionalProperties = refed schema
- additionalProperties omitted -> {} (any type)
Because that would comply with the Openapi v2 and v3 specifications. Before that though, we should aim for consistency of results when parsing v2 and v3 specs.

This fix is applied to the v1 branch, and should rev roll it to a new version, maybe 1.1.0 because this may be a breaking change for legacy usages of additionalProperties which did not handle true/false/and omitted correctly.
We can't rev roll to 2.X.X because that is used by the current master branch which handles v3 specs and uses this branch for v2 spec parsing before conversion to v3 specs.

@sebastien-rosset 
@gracekarina 
@hkosova


spacether:
Note: a couple of tests are not passing yet; I am working on them

modules/swagger-parser/src/main/java/io/swagger/parser/util/SwaggerDeserializer.java
```diff
@@ -819,6 +820,10 @@ public Model definition(ObjectNode node, String location, ParseResult result) {
             JsonNode ap = node.get("additionalProperties");
             if (ap != null && ap.getNodeType().equals(JsonNodeType.OBJECT)) {
                 impl.setAdditionalProperties(Json.mapper().convertValue(ap, Property.class));
+            } else if ("object".equalsIgnoreCase(type) && (ap == null || "true".equalsIgnoreCase((String) ap.asText()))) {
```
spacether, line 823: Should we set additionalProperties true for composed schemas also?
What does the master branch of swagger parser do?

spacether:
@gracekarina can you help debug these failures?
[This other PR that I just opened](v1 PR with comment change, shows CI failing for v1 branch) with no functional changes on the v1 branch also has these same 2 test failures

spacether:
> @gracekarina can you help debug these failures?

Never mind; I was able to fix them on my own.
@gracekarina all tests pass. Can we get a review on this PR?

modules/swagger-parser/src/test/java/io/swagger/parser/NetworkReferenceTest.java
```diff
@@ -62,12 +65,43 @@
...
+                        "            $ref: 'http://petstore.swagger.io/v2/swagger.json#/definitions/Pet'";
+
+        SwaggerDeserializationResult result = new SwaggerParser().readWithInfo(spec);
+
+        Swagger swagger = result.getSwagger();
+        Property property = swagger.getPath("/foo").getGet().getResponses().get("200").getSchema();
+        assertNotNull(property);
+        assertTrue(property instanceof RefProperty);
+    }
+
```
spacether, line 104: Moved + response mocked to get test to pass

modules/swagger-parser/src/test/java/io/swagger/parser/NetworkReferenceTest.java
```diff
@@ -248,4 +282,43 @@ public void testIssue411() throws Exception {
...
+        assertNotNull(swagger.getPath("/store/order").getPost().getParameters());
+        assertTrue(swagger.getPath("/store/order").getPost().getParameters().size() == 1);
+        assertTrue(swagger.getPath("/store/order").getPost().getParameters().get(0) instanceof BodyParameter);
+        assertNotNull(((BodyParameter)swagger.getPath("/store/order").getPost().getParameters().get(0)).getSchema());
+        assertTrue(((BodyParameter)swagger.getPath("/store/order").getPost().getParameters().get(0)).getSchema() instanceof RefModel);
+        assertTrue(((RefModel)((BodyParameter)swagger.getPath("/store/order").getPost().getParameters().get(0)).getSchema()).getSimpleRef().equals("Order"));
+
+        assertTrue(swagger.getDefinitions().get("Order") instanceof ModelImpl);
+        assertTrue(swagger.getDefinitions().get("Order").getProperties().size() == 6);
+    }
```
spacether, line 323: Moved + response mocked to get test to pass

modules/swagger-parser/src/test/java/io/swagger/parser/SwaggerParserTest.java
```diff
@@ -1648,4 +1648,36 @@ public void testIssue913() {
...
+        schema = (ModelImpl) swagger.getDefinitions().get("AdditionalPropertiesIsString");
+        assertNotNull(schema);
+        assertNotNull(schema.getAdditionalProperties());
+        assertEquals(schema.getAdditionalProperties().getType(), "string");
+
+        schema = (ModelImpl) swagger.getDefinitions().get("AdditionalPropertiesIsRef");
+        assertNotNull(schema);
+        assertNotNull(schema.getAdditionalProperties());
+        assertEquals(schema.getAdditionalProperties().getType(), "ref");
+    }
```
spacether, line 1682: these are our additionalProperties tests

modules/swagger-parser/src/test/java/io/swagger/parser/NetworkReferenceTest.java
gracekarina: I have a question why the need to move the test from one class to another?

spacether: It was easier to do it this way. All of our classes that we need to attach the fake request already exist in the file that I moved it in to. If I kept them where they were then I would need to add new import and setup lines for mocking the request for just those two tests in two separate files. We already have that set up code in the file that I moved them into.

spacether:
Converting to draft while I am making these changes:
TODO, update to:
```
additionalProperties = true -> Boolean.TRUE
additionalProperties = false -> Boolean.TRUE
additionalProperties = string schema
additionalProperties = refed schema
additionalProperties omitted -> null
This will bring our results in line with the v3 spec parsing.
```

spacether:
Making the TODO changes was too arduous so I stopped working on this.

spacether:
Closing this because there doesn't appear to be interest from the swagger-parser team in merging it and feedback in https://github.com/swagger-api/swagger-parser/issues/1369 is that the v2 openapi spec definition of additionalProperties is not consistent with that the openapi spec authors intended.

Issue #1419, opened 2020-08-17
Parser is throwing error for a valid OpenAPI File


SCSreekanth:
The attached content is valid but the parser is throwing error " attribute components.schemas.Vehicle.enum is not of type `value`"
```json
{
  "openapi": "3.0.0",
  "servers": [{
      "description": "Production environment",
      "url": "https://driver-vehicle-licensing.api.gov.uk/vehicle-enquiry"
    },
    {
      "description": "Test environment",
      "url": "https://uat.driver-vehicle-licensing.api.gov.uk/vehicle-enquiry"
    }
  ],
  "info": {
    "contact": {
      "email": "DvlaAPIAccess@dvla.gov.uk",
      "name": "DVLA API Team",
      "x-twitter": "GDSTeam"
    },
    "description": "Interface specification for the DVLA Vehicle Enquiry API",
    "title": "Vehicle Enquiry",
    "version": "1.1.0",
    "x-apisguru-categories": [
      "open_data"
    ],
    "x-logo": {
      "url": "https://api.apis.guru/v2/cache/logo/https_twitter.com_GDSTeam_profile_image.png"
    },
    "x-origin": [{
      "format": "openapi",
      "url": "https://developer-portal.driver-vehicle-licensing.api.gov.uk/apis/vehicle-enquiry-service/v1.1.0-vehicle-enquiry-service.json",
      "version": "3.0"
    }],
    "x-preferred": true,
    "x-providerName": "api.gov.uk",
    "x-serviceName": "vehicle-enquiry"
  },
  "tags": [{
    "name": "vehicle"
  }],
  "paths": {
    "/v1/vehicles": {
      "post": {
        "description": "Returns vehicle details based on registration number",
        "operationId": "getVehicleDetailsByRegistrationNumber",
        "parameters": [{
            "description": "Client Specific API Key",
            "in": "header",
            "name": "x-api-key",
            "required": true,
            "schema": {
              "type": "string"
            }
          },
          {
            "description": "Consumer Correlation ID",
            "in": "header",
            "name": "X-Correlation-Id",
            "required": false,
            "schema": {
              "type": "string"
            }
          }
        ],
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/Vehicle"
              }
            }
          },
          "description": "Registration number of the vehicle to find details for",
          "required": true
        },
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
[...]

PR #1453, opened 2020-10-08
Fix for issue-1419

r-sreesaran:
Fix for issue-1419.
The issue here is that when enum field is having an array value, It is getting logged as error.
This was fixed as the missing condition was handled. The value is currently saved as string.


gracekarina (approved):
Thanks @r-sreesaran

kerrykimbrough:
I think this fix is incorrect and should not be released. See comments on issue #1419

gracekarina:
reverting the merge please see comments on issue #1419

PR #1456, opened 2020-10-09
fix NPE when parsing V2 with OpenAPIV3Parser and resolveFully option

### LINKED ISSUE #1455: OpenAPIParser - NPE when parsing swagger V2 with resolveFully option
When I parse a swagger v2 descriptor with the OpenAPIParser and the `resolveFully` option set to true, I get this NPE:
```
WARN  io.swagger.v3.parser.OpenAPIV3Parser - Exception while resolving:
java.lang.NullPointerException: null
    at io.swagger.v3.parser.util.ResolverFully.resolveFully(ResolverFully.java:60)
    at io.swagger.v3.parser.OpenAPIV3Parser.resolve(OpenAPIV3Parser.java:174)
    at io.swagger.v3.parser.OpenAPIV3Parser.readContents(OpenAPIV3Parser.java:153)
    at io.swagger.v3.parser.OpenAPIV3Parser.readLocation(OpenAPIV3Parser.java:89)
    at io.swagger.parser.OpenAPIParser.readLocation(OpenAPIParser.java:16)
```

the root cause is here:
https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java#L166
```java
private SwaggerParseResult resolve(SwaggerParseResult result, List<AuthorizationValue> auth, ParseOptions options,
            String location) {
        try {
            if (options != null) {
                if (options.isResolve() || options.isResolveFully()) {
                    result.setOpenAPI(new OpenAPIResolver(result.getOpenAPI(), emptyListIfNull(auth),
                            location).resolve());
                    if (options.isResolveFully()) {
                        new ResolverFully(options.isResolveCombinators()).resolveFully(result.getOpenAPI());
                    }
                }
...
```
IMHO the `resolve` method should return the `result` parameter if `result.getOpenAPI()` is null (Which is the case when parsing a swagger V2 in OpenApiV3Parser)

phiz71:
Avoid to resolve openAPI if it is null
Fixes #1455

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
@@ -150,7 +150,10 @@ public class OpenAPIV3Parser implements SwaggerParserExtension {
             final ObjectMapper mapper = getRightMapper(swaggerAsString);
             final JsonNode rootNode = mapper.readTree(swaggerAsString);
             final SwaggerParseResult result = parseJsonNode(location, rootNode);
-            return resolve(result, auth, options, location);
+            if (result.getOpenAPI() != null) {
+                return resolve(result, auth, options, location);
+            }
+            return result;
         } catch (JsonProcessingException e) {
             LOGGER.warn("Exception while parsing:", e);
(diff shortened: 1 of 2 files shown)


gracekarina (approved):
Thanks @phiz71

Issue #1457, opened 2020-10-10
Error when setting multipleOf to a value between 0 and 1


feliperuiz:
I was getting a NullPointerException when validating a spec using `swagger-parser@2.0.22`, turns out it would happen because I had a `multipleOf` property set to `0.5`, which triggered an `else` branch that works on a `null` variable. The only limitation for multiples, as mentioned in [Swagger's own docs](https://swagger.io/docs/specification/data-models/data-types/), is that:
> The value of `multipleOf` must be a positive number

A simple schema such as the one below is enough to reproduce the error:
```yaml
openapi: 3.0.0
info:
  version: 0.0.0
  title: test
components:
  schemas:
    Test:
      type: object
      properties:
        decimal_value:
          type: number
          multipleOf: 0.3
```

I already have a patch in place and will submit a PR, if you folks don't mind it.

gracekarina:
fixed by #1458

PR #1458, opened 2020-10-10
Handle multipleOf between 0 and 1

feliperuiz:
This pull request fixes handling of `multipleOf` between 0 and 1, as specified in #1457. I've also added a test case for this specific scenario.

The existing code would call `intValue()`, which involves rounding. When that would happen, low values would round to 0, failing the ensuing test.

P.S.: Getting to the `else` branch trigger a `NullPointerException` because `result` is `null`. I haven't tried to fix that issue as I believe it would be a completely different beast.


kerrykimbrough:
👍

gracekarina:
@feliperuiz thanks for the PR!

Issue #1479, opened 2020-11-05
v1-branch: Replace outdated fge-dependencies with new java-json-tools dependencies


sdoeringNew:
The work of fge found a new owner / maintainer. All fge-dependencies have been moved to java-json-tools groupId.

gracekarina:
Hi @sdoeringNew, thanks for the PR, where can I read about that change? I have looked at the fge project and I can't find any mention to the java-json-tools .

sdoeringNew:
You can't find it there anymore.

Look at: https://github.com/fge/json-schema-validator
It redirects to the new maintainer: https://github.com/java-json-tools/json-schema-validator

And you can see fge's last update on the README.md looking for a new maintainer: https://github.com/java-json-tools/json-schema-validator/commit/fd781ef2

PR #1480, opened 2020-11-05
Replace fge-dependencies with their java-json-tools replacements

sdoeringNew:
This PR handles issue #1479.


gracekarina:
thanks @sdoeringNew

PR #1482, opened 2020-11-05
fixed bug related to nested definitions within subfolder

volkodava:
In this pull request we have fix for nested references for external references within subfolder.


gracekarina:
thanks @volkodava

PR #1490, opened 2020-11-13
OpenAPIDeserializer.getSchema: Handle "enum" correctly when schema type is "array" or "object"

kerrykimbrough:
Fixes issue #1419 

For a schema of type `array`, each element of the `enum` array must be an array.  But an enumerated array may contain values that don't have a native Java representation (for example, "object" values). Therefore, an enumerated array element is represented by an `ArrayNode` value, which contains the fully-parsed representation of the enumerated value.

Similarly, for a schema of type `object`, each element of the `enum` array must be an object. But an object value doesn't have a well-defined native Java representation. Therefore, an enumerated object element is represented by an `ObjectNode`.

But in any case, a null element of an `enum` array is represented by a Java `null` value, which is consistent with `enum` handling for all other schema types.


gracekarina:
thanks @kerrykimbrough

PR #1502, opened 2020-12-02
Fix issue #1501: External schema name is correctly constructed

### LINKED ISSUE #1501: Component schemas with same name are ignored after the second one
When several referenced schemas have the same name but different definitions, only the first two are created in components/schemas, the rest are ignored.
### base.yaml
```
openapi: 3.0.0
paths:
  /endpoint:
    post:
      summary: test summary
      description: test desc
      requestBody:
        required: true
        content:
          application/json:
            schema:
              oneOf:
                - $ref: 'config.yaml#/POST.type1/body'
                - $ref: 'config.yaml#/POST.type2/body'
                - $ref: 'config.yaml#/POST.type3/body'
                - $ref: 'config.yaml#/POST.type4/body'
      responses:
        '200':
          description: OK
```
### config.yaml
```
POST.type1:
  body:
    type: object
    additionalProperties: false
    properties:
      prop1:
        description: prop1
        type: boolean
POST.type2:
  body:
    type: object
    additionalProperties: false
    properties:
      prop2:
        description: prop2
        type: boolean
POST.type3:
  body:
    type: object
    additionalProperties: false
    properties:
      prop3:
        description: prop3
        type: boolean
POST.type4:
  body:
    type: object
    additionalProperties: false
    properties:
      prop3:
        description: prop3
        type: boolean
```
### main.java
```
import io.swagger.v3.oas.models.OpenAPI;
import io.swagger.v3.parser.core.models.ParseOptions;

public class Main {

    public static void main(String[] args) {
        options.setResolve(true);
        OpenAPI openAPI = new OpenAPIV3Parser().read("base.yaml", null, options);
        System.out.println(openAPI.getComponents().getSchemas().keySet());
    }
}
```
### Result
```
[body, body_2]
```
### Expected result
```
[body, body_1, body_2]
```
Note: In this case 3 schemas are expected since `POST.type3.body` and `POST.type4.body` have the same definition.

BabisK:
Calculation of schema name in external ref processor takes into account that more than 2 schemas may have the same name.
The counting of schemas starts from `1` instead of `2`.

Note: the warn level of the logger in line 62 seemed too high for a normal operation so I lowered it. Let me know if you think it was `warn` for an abnormal operation.

Closes #1501

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -49,6 +49,23 @@ public final class ExternalRefProcessor {
         this.openAPI = openAPI;
     }
 
+    private String finalNameRec(Map<String, Schema> schemas, String possiblyConflictingDefinitionName, Schema newScema,
+        int iteration) {
+        String tryName =
+            iteration == 0 ? possiblyConflictingDefinitionName : possiblyConflictingDefinitionName + "_" + iteration;
+        Schema existingModel = schemas.get(tryName);
+        if (existingModel != null) {
+            if (existingModel.get$ref() != null) {
+                // use the new model
+                existingModel = null;
+            } else if (!newScema.equals(existingModel)) {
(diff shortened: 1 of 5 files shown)


gracekarina (approved):
thanks @BabisK.

PR #1536, opened 2021-03-11
refactor: remove repeated else-if

andrsdt:
This PR does the following:
* Removing an unreachable branch with a duplicated condition. The block inside is the same as covered above.


gracekarina:
Thanks @andrsdt

PR #1540, opened 2021-03-23
Fix the problem that the content of extensions cannot be parsed

Bulls1986:
Hello,bro.It seems that the content of extensions under the node is not parsed here,resulting in the extensions content of all nodes not being parsed,so here is added to take the extensions node and parse the content under the node.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -72,3110 +72,3176 @@ import java.util.stream.Stream;
 
 public class OpenAPIDeserializer {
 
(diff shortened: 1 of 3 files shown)


Bulls1986:
This problem will appear in Json format.

gracekarina:
Hi, thanks for the PR, may you please commit the specific change you are making, and a specific test to corroborate the fix, thanks.

Bulls1986:
The new logic is that after the original extensions are parsed, if the content is not parsed, try to parse it again in Json format. I have added the unit test.

gracekarina:
thanks @Bulls1986

Msimplehub:
How  to fix it in 2.0.0-rc1.
and the dependency is :
<dependency>
    <groupId>io.swagger</groupId>
    <artifactId>swagger-parser</artifactId>
    <version>2.0.0-rc1</version>
</dependency>

PR #1541, opened 2021-03-26
fix schemas parse inconsistencies

GabrielCastro:
schemas an of inline responses are parsed different from ones in definitions.

This manifests it's self as a bug in the generator: https://github.com/OpenAPITools/openapi-generator/issues/9086 where `additionalProperties` behaves differently if it's in a response in-line or in definitions.


GabrielCastro:
@frantuma 
@gracekarina

amjmhs:
isn't this the same issue as #1552 and according PR #1554 ?

GabrielCastro:
> isn't this the same issue as #1552 and according PR #1554 ?

Yes, it is

GabrielCastro:
It would be great to have either fix merged
cc: @frantuma @gracekarina

gracekarina:
thanks for the PR and adding the test! @GabrielCastro

PR #1542, opened 2021-03-26
external $ref without / caused exception

### LINKED ISSUE #1543: External $ref without '/' causes exception
$ref("another_file.json#/something") causes StringIndexOutOfBounds, becuase no '/' is found in left part of reference. In addition, $ref("../../a_third_file.json#/something") also causes StringIndexOutOfBounds on Windows, because the code searches for File.separatorChar (= '\\') as path separator.

kaewberg:
Fixes #1543
... subtring(0, indexOf ... causes StringIndexOutOfBounds if the char searched for is missing. This meant that refs like $ref("another_file.json#/thing") could not be parsed. In addition, File.sparatorChar was used to find separators in directory part of ref, causing most every parsing attempt to fail on windows. We should probably require "/" to be used as path separator.


gracekarina:
Hi, @kaewberg thanks for the PR, can you please add the spec that is causing this error?, or even better if you can add a test to the PR will be great!. Thanks

gracekarina:
thanks @kaewberg

Issue #1459, opened 2020-10-10
Swagger Parser doesn't handle relative file $refs in paths according to RFC3986


feliperuiz:
When traversing `$ref`s to local files, Swagger Parser breaks at the first level of indirection due to not properly adjusting the base path of subsequent `$ref`s. An example might better illustrate it, so let's take the following tree:

```
├─ openapi
│  ├─ entrypoint.yaml
│  └─ common-params.yaml
│
└─ module-1
   └─ openapi
      ├─ paths
      │  └─ path-1.yaml
      │
      └─ schemas
         └─ schema-1.yaml
```

We also have the following `$ref`s:
<table>
<thead>
  <tr>
    <th>Spec file</th>
    <th>Has relative <code>$ref</code> to</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td nowrap><code>openapi/entrypoint.yaml</code></td>
    <td nowrap><code>../module-1/openapi/paths/path-1.yaml#/paths/~1path-1</code></td>
  </tr>
  <tr>
    <td nowrap rowspan="2" align="center"><code>module-1/openapi/path-1.yaml</code></td>
    <td nowrap><code>../../../openapi/common-params.yaml#/components/parameters/param-1</code></td>
  </tr>
  <tr>
    <td nowrap><code>../schemas/schema-1.yaml#/components/schemas/schema-1</code></td>
  </tr>
</tbody>
</table>

Trying to parse this with the following snippet
```java
var options = new ParseOptions();
options.setResolve(true);

var result = new OpenAPIParser().readLocation("openapi/entrypoint.yaml", null, options);
```
results in the error message below
```
Unable to load RELATIVE ref: ../../../openapi/common-params.yaml
```

I've done some investigation, it seems that once the parser resolves the `$ref` to `../module-1/openapi/paths/path-1.yaml#/paths/~1path-1`, all of its contents are "dumped" to the result object being built, as if they came from the original `entrypoint.yaml` file. Local `$ref`s within it are correctly updated to account for this change, but relative ones are not, meaning the parser will try to look for `../../../openapi/common-params.yaml` starting from `openapi/entrypoint.yaml`. It could either look for `../../../openapi/common-params.yaml` starting from `module-1/openapi/paths/` or adjust the `$ref` to account for using the wrong base path.

[...]

PR #1557, opened 2021-04-26
fix handling of relative references

johanra:
This PR includes a fix & testcase for this [issue] (https://github.com/swagger-api/swagger-parser/issues/1459)

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/PathsProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/PathsProcessor.java
@@ -65,7 +65,7 @@ public class PathsProcessor {
                 String pathRef = pathItem.get$ref().split("#")[0];
 
                 if (resolvedPath != null) {
-                    updateLocalRefs(resolvedPath, pathRef);
+                    updateRefs(resolvedPath, pathRef);
                     //we need to put the resolved path into swagger object
                     openAPI.path(pathStr, resolvedPath);
                     pathItem = resolvedPath;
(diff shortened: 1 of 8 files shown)


gracekarina:
thanks @johanra

olesiathoms-wk:
@gracekarina Do you know when this fix will be released?

slimane-deb:
@johanra Looking forward to it. Thumbs up

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/PathsProcessor.java
```diff
@@ -287,6 +277,27 @@ protected boolean isLocalRef(String ref) {
         }
         return false;
     }
+    
+    protected boolean isAbsoluteRef(String ref) {
+    	 if(ref.startsWith("./")) {
+             return true;
+         }
```
feliperuiz, line 284: I'm sorry for asking a long time after this was merged, but wouldn't this make it a relative ref instead of an absolute one? Wouldn't it need to start with a `<scheme>:` to be considered [absolute](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3)?

johanra: You are right this is not correct.  
My intention was to have the isAbsoluteRef method match anything that is not relative.
The code should  then look like this:
protected boolean isAbsoluteRef(String ref) { return !ref.startsWith("."); ​}

protected String computeRelativeRef(String ref, String prefix) {
if(ref.startsWith("./")) {
​return ref;
​}
int iIdxOfSlash = prefix.lastIndexOf('/');
if(iIdxOfSlash != -1) {
return prefix.substring(0, iIdxOfSlash+1) + ref;
}
return prefix + ref;
​}

Issue #1562, opened 2021-05-12
hundreds of inline 'body123' types when using large APIs (e.g. msgraph)


ivy-rew:
using the parser with flatten option+ codegen for generating java pojos we get hundreds of Body123 pojos.
Which makes them complicate to use in java. ... was it Body57 or Body59 that I need to send? 

This happens to the extremes if I use ms-graph apis which contain many inline body types.
https://graphexplorerapi.azurewebsites.net/openapi?tags=me.user,me.calendar,users.calendar,me.message,me.Actions,me.todo&openApiVersion=3&graphVersion=v1.0&format=json&style=PowerShell

However, we could easily make things simpler and better to read, by adding path parts to the 'body' name. And therefore avoid too many instances of the same type with number postfixes.

ivy-rew:
Thanks for accepting the PR @gracekarina . So I'm happily closing this PR and switch to parser 2.0.26, once the final release is available. :+1:

PR #1563, opened 2021-05-12
Bug/1562 path body

ivy-rew:
expressive 'body' schemas that make them much easier to map mentally to the related path.

https://github.com/swagger-api/swagger-parser/issues/1562


gracekarina:
thanks @ivy-rew. This looks great!

Issue #1553, opened 2021-04-11
swagger-parser-v3 fails to generate spec on relative server urls on Windows


jhannes:
After a change in OpenAPIDeserializer, swagger-parser-v3 fails to generate any model for a spec with a relative server url when running on Windows.

The message recorded in result is "begin 0, end -1, length 3", quite unhelpfully, while the result of `new OpenAPIParser().readLocation(inputSpec, authorizationValues, options)` is just result.openAPI and `result.messages` both being null

According to the [OpenAPI spec](https://swagger.io/docs/specification/api-host-and-base-path/) relative base paths are supported by the OpenAPI standard. Relative URLs are of course very useful when developing a backend-for-frontend API. At any rate, if swagger-parser was to not support it, users deserve a better error message.

Here is a minimum OpenAPI spec to reproduce the problem:

```yaml
openapi: 3.0.2
info:
  title: Sample API
  description: A small example to demonstrate individual problems
  version: 0.1.9
servers:
  - url: /v1
    description: Server
paths:
  /pets:
    patch:
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Pet'
      responses:
        '200':
          description: Updated
components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
        pet_type:
          type: string
        name:
          type: string
        birth_date:
          type: string
          format: date
```

As of now, I'm executing the parser through openapi-generator, so I don't have a targeted test case or pull request, but can work on this if you want.

## Cause:

[Trying to substitute server variable values when there are no variables](https://github.com/swagger-api/swagger-parser/blob/41578a6f2ecce74bbdca6525aa664b7e9c2bd2fc/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java#L426) causes a `StringIndexOutOfBoundsException`. This line should've checked on `value.indexOf("{")` before doing `substring`. Additionally, the code does not support multiple template variables as [required by the spec](https://swagger.io/docs/specification/api-host-and-base-path/). Please not that the code here uses tabs and spaces inconsistently and looks bad on github.

The substitution only happens on URIParseException, which is caused by Windows-style paths with "\" instead of "/".

[...]

PR #1565, opened 2021-05-19
#1553 fails to generate spec on relative server urls on Windows

jhannes:
When parsing API specs with relative server URLs when the path to the spec is specified with Windows directory separators (\) instead of UNIX (/), swagger-parser fails to generate OpenAPI object with no error message.

This is the cause of #1553 and most likely of https://github.com/OpenAPITools/openapi-generator/issues/8266 as well


gracekarina (approved):
Thanks a lot for this PR.

PR #1616, opened 2021-10-06
implementation of 3.1 support deserialization and validation of non JsonSchema fields

gracekarina:
implementation of 3.1 support deserialization and validation of non JsonSchema fields

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -266,10 +266,10 @@ public class OpenAPIDeserializer {
 	private final Set<String> operationIDs = new HashSet<>();
 
 	public SwaggerParseResult deserialize(JsonNode rootNode) {
-		return deserialize(rootNode, null);
+		return deserialize(rootNode, null, false);
 	}
 
-	public SwaggerParseResult deserialize(JsonNode rootNode, String path) {
+	public SwaggerParseResult deserialize(JsonNode rootNode, String path, boolean isOaiAuthor) {
 		basePath = path;
 		this.rootNode = rootNode;
 		rootMap = new ObjectMapper().convertValue(rootNode, Map.class);
@@ -277,6 +277,7 @@ public class OpenAPIDeserializer {
 		try {
 
 			ParseResult rootParse = new ParseResult();
+			rootParse.setOaiAuthor(isOaiAuthor);
 			OpenAPI api = parseRoot(rootNode, rootParse, path);
 			result.openapi31(rootParse.isOpenapi31());
 			result.setOpenAPI(api);
@@ -378,7 +379,11 @@ public class OpenAPIDeserializer {
 			if (result.isOpenapi31()) {
 				value = getString("jsonSchemaDialect", rootNode, false, location, result);
 				if (value != null) {
-					openAPI.setJsonSchemaDialect(value);
+					if (isValidURL(value)) {
+						openAPI.setJsonSchemaDialect(value);
+					}else{
+						result.warning(location,"jsonSchemaDialect. Invalid url: " + value);
+					}
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -369,12 +375,25 @@ public OpenAPI parseRoot(JsonNode node, ParseResult result, String path) {
 				openAPI.setExtensions(extensions);
 			}
 
+			if (result.isOpenapi31()) {
+				value = getString("jsonSchemaDialect", rootNode, false, location, result);
```
frantuma, line 379: it makes sense to check if the string is a URL, according to spec

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -386,6 +405,15 @@ public OpenAPI parseRoot(JsonNode node, ParseResult result, String path) {
 		return openAPI;
 	}
 
+	private void validateReservedKeywords(Map<String, Set<String>> specKeys, String key, String location, ParseResult result) {
+		if(result.isOpenapi31() && specKeys.get("RESERVED_KEYWORDS").stream()
```
frantuma, line 409: better to add a new boolean flag in result, like `oaiInternal` or `oaiAuthor` to allow for any key prefix if the author is indeed OAI

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -2153,9 +2195,9 @@ public SecurityScheme getSecurityScheme(ObjectNode node, String location, ParseR
 		}
 
 		boolean descriptionRequired, bearerFormatRequired, nameRequired, inRequired, schemeRequired, flowsRequired,
-				openIdConnectRequired;
+				openIdConnectRequired, mutualTLSRequired;
```
frantuma, line 2198: mutualTLSRequired not used, what is this about?

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -2397,6 +2442,7 @@ public Discriminator getDiscriminator(ObjectNode node, String location, ParseRes
 					if (extensions != null && extensions.size() > 0) {
 						discriminator.setExtensions(extensions);
 					}
+					//validateReservedKeywords(keys,key, location, result);
```
frantuma, line 2445: why is this commented?

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -4012,6 +4069,11 @@ public boolean isOpenapi31() {
 				String message = "attribute " + location + l.key + " is repeated";
 				messages.add(message);
 			}
+			for (Location l : reserved) {
+				String location = l.location.equals("") ? "" : l.location + ".";
+				String message = "attribute " + location + l.key + " is reserved by The OpenAPI Iniciative";
```
frantuma, line 4074: Typo: `Iniciative` -> `Initiative`

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -2821,6 +2867,7 @@ at the moment path passed as string (basePath) from upper components can be both
 			if (!specKeys.get("SCHEMA_KEYS").contains(key) && !key.startsWith("x-")) {
 				result.extra(location, key, node.get(key));
 			}
+			validateReservedKeywords(specKeys, key, location, result);
```
frantuma, line 2870: This is handling 3.0, so not applicable. Also I believe this restriction is not applied to Json Schema also in 3.1

frantuma (changes requested):
On top of comments, please add a test demonstrating that a spec with all new constructs parses ok, with the fields populated

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
gracekarina: I'll remove the validation from there but I think in the method we'll handle 3.1 schema(JsonSchema) we may find also extensions with `x-` or without.  
`This object MAY be extended with Specification Extensions, though as noted, additional properties MAY omit the x- prefix within this object.`

frantuma:
@gracekarina if I get this correctly, this PR is replaced by #1617 and should therefore be closed?

Issue #1621, opened 2021-10-26
OpenAPIParser is resolving the wrong schema for an operation


simondean:
See the following repo for a minimal reproduction of this issue: 

* https://github.com/simondean/swagger-request-validator-filename-issue
* https://github.com/simondean/swagger-request-validator-filename-issue/blob/main/src/test/java/IssueDemoTest.java
* https://github.com/simondean/swagger-request-validator-filename-issue/blob/main/example-openapi.yaml

Here's an example OpenAPI spec that demonstrates the issue: 

https://github.com/simondean/swagger-request-validator-filename-issue/blob/main/example-openapi.yaml
```yaml
openapi: 3.0.1
info:
  title: Example OpenAPI spec
  version: 0.0.1
paths:
  '/example':
    post:
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/ExamplePost'
      responses:
        204:
          description: No content
components:
  schemas:
    # Renaming this key from `Example` to `ExampleGet` stops the issue from occurring
    # Removing this key stops the issue from occurring
    Example:
      $ref: example.get.yaml
    ExamplePost:
      $ref: example.post.yaml
```

The comments in the code snippet show how the issue only happens when the unused schema is called `Example`.  Rename it to `ExampleGet` or delete it, and the issue stops.  The word `Example` is unimportant, other than it seems to need to be similar to the other schema that is used.  Both schemes could be renamed to `Anything` and `AnythingPost` and you would still get the issue

## Expected Behaviour

The following assertion should pass: 

```java
        assertEquals("POST Example", api.getPaths()
                .get("/example")
                .getPost()
                .getRequestBody()
                .getContent()
                .get("application/json")
                .getSchema()
                .getTitle());
```

## Actual Behaviour

Swagger Parser loads the wrong request body schema for the POST operation.  It's request body description is set to `GET Example` when it should be `POST Example`:

```
expected: <POST Example> but was: <GET Example>
```

PR #1622, opened 2021-10-28
Fix for issue-1621

ymohdriz:
Hi team,

This PR is to solve issue #1621  (OpenAPIParser is resolving the wrong schema for an operation)
This issue happens when the referenced file name contains more than 1 dot.

Please review the PR and merge the same.

Thanks,
Mohammed Rizwan


gracekarina (approved):
thanks I will merged this now

PR #1624, opened 2021-11-02
Fix for issue 1601

ymohdriz:
Parser does not validate GET request with request body


gracekarina (changes requested):
hi @ymohdriz thanks for this PR, but sadly I cannot merged as is, because it will introduce a breaking change for those who use parser and do not expect that message validation. I will keep this in the radar for when we implement a different message handler, so when something is a warning won't break the validation of parser.

ymohdriz:
Hi @gracekarina,

Make sense. 
What if we can add this warning only if it's OAI version >= 3.0.3 so that it won't affect the backward compatibility. Your thoughts, please?

Thanks,
Mohammed Rizwan

gracekarina:
> Hi @gracekarina,
> 
> Make sense. What if we can add this warning only if it's OAI version >= 3.0.3 so that it won't affect the backward compatibility. Your thoughts, please?
> 
> Thanks, Mohammed Rizwan

Let's wait until we do some improvements at the way parser warns it will be a full change. thanks

Issue #1524, opened 2021-01-26
Incompatible with Jackson 2.12


sirianni:
Using this library with Jackson 2.12 fails with the following error:
```
ERROR Java 8 date/time type `java.time.OffsetDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: io.swagger.v3.oas.models.media.ComposedSchema["allOf"]->java.util.ArrayList[0]->io.swagger.v3.oas.models.media.ObjectSchema["properties"]->java.util.LinkedHashMap["metadata"]->io.swagger.v3.oas.models.media.Schema["properties"]->java.util.LinkedHashMap["created_at"]->io.swagger.v3.oas.models.media.DateTimeSchema["example"])
```

In https://github.com/FasterXML/jackson-databind/issues/2776 it appears that Jackson 2.12 no longer handles Java 8 date types without `JavaTimeModule` being explicitly registered.

frantuma:
fixed in #1625

PR #1625, opened 2021-11-10
add JavaTimeModule to Jackson ObjectMapper

pjfanning:
Relates to https://github.com/swagger-api/swagger-parser/issues/1524 - also consistent with swagger-core ObjectMapperFactory that adds JavaTimeModule by default.


frantuma:
Thanks!

PR #1627, opened 2021-11-17
when the same component used in two different places in two different apis then parser fails

ptrsd:
Hi, I'm getting ClassCastException when the same parameter from $ref is used in different places in different APIs. In this PR I've added a test case for this problem.

Exception:
```
java.lang.ClassCastException: Cannot cast io.swagger.v3.oas.models.headers.Header to io.swagger.v3.oas.models.parameters.Parameter
	at java.base/java.lang.Class.cast(Class.java:3816)
	at io.swagger.v3.parser.ResolverCache.loadRef(ResolverCache.java:116)
	at io.swagger.v3.parser.processors.ParameterProcessor.processParameters(ParameterProcessor.java:86)
	at io.swagger.v3.parser.processors.OperationProcessor.processOperation(OperationProcessor.java:39)
	at io.swagger.v3.parser.processors.PathsProcessor.processPaths(PathsProcessor.java:84)
	at io.swagger.v3.parser.OpenAPIResolver.resolve(OpenAPIResolver.java:49)
	at io.swagger.v3.parser.OpenAPIV3Parser.resolve(OpenAPIV3Parser.java:183)
	at io.swagger.v3.parser.OpenAPIV3Parser.readContents(OpenAPIV3Parser.java:162)
	at io.swagger.v3.parser.OpenAPIV3Parser.readContents(OpenAPIV3Parser.java:99)
```

### DIFF
--- a/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java
+++ b/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java
@@ -639,6 +639,52 @@ public class OpenAPIV3ParserTest {
     }
 
     @Test
+    public void testSameComponentInRequestAndResponseHeaders() throws IOException {
+        String pathFile = FileUtils.readFileToString(new File("src/test/resources/same-header-different-place-domain.yaml"), "UTF-8");
+        WireMock.stubFor(get(urlPathMatching("/issue-domain"))
+                .willReturn(aResponse()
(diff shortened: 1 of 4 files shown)


ewaostrowska:
Thanks for reporting this @ptrsd . I agree that the parser should not throw a raw `ClassCastException`, so there is a real issue here.

However, I don't think the tests in this PR describe valid OpenAPI behavior and I would also prefer to have it reported as an issue if this problem still occurs as I am unable to merge the PR anyway if the tests are failing. 

I am closing this PR. If the issue persists please report it as an issue according to the current template

Issue #1636, opened 2022-01-04
Schema without type array is ignoring items in Responses 


SCSreekanth:
Hi Team,

Please find below the swagger content in which for a response of the operation, the schema is set as items but without type array. The parser doesn't throw any error here but the response model that is generated has ignored the items. Please correct this.

```
swagger: '2.0'
info:
  description: 'This is a TEST.'
  version: 1.0.0
  title: Test
host: www.abc.com
basePath: /api
schemes:
  - http
paths:
  /test:
    get:
      summary: Test
      description: 'test'
      operationId: test
      responses:
        '200':
          schema:
            items:
              $ref: '#/definitions/myResponse'
          description: myResponse
definitions:
  myResponse:
    type: object
    properties:
      id:
        type: integer
        format: `int64'
```

Thanks,
Sree

choiks14:
same here

gracekarina:
this issue was addressed in this two prs: Branchs affected:
core 1.5:  https://github.com/swagger-api/swagger-core/pull/4109
parser v1: https://github.com/swagger-api/swagger-parser/pull/1642/

SCSreekanth:
@gracekarina - Now that this issue is closed, could you please let me know the parser version in which this issue will be fixed? Thanks
Sree

gracekarina:
@SCSreekanth we have a release scheduled in a couple of weeks.

SCSreekanth:
Any update on the jar release and its version.?

gracekarina:
Hi @SCSreekanth  parser was release last week, this fix is include in the last version 2.0.30.

PR #1642, opened 2022-01-27
fix for #1636

gracekarina:
this PR add support for items in all Schema types in OpenAPI 2.0 #1636 
depends on: https://github.com/swagger-api/swagger-core/pull/4109


frantuma (changes requested):
I would drop this and replaced with what mentioned in [related core PR comment](https://github.com/swagger-api/swagger-core/pull/4109#pullrequestreview-865941080)

Issue #1646, opened 2022-02-02
Invalid parameter "content" not recognized


kerrykimbrough:
Here's how the OpenAPI spec defines the "content" field for a Parameter Object:

> A map containing the representations for the parameter. The key is the media type and the value describes it. The map MUST only contain one entry.

Accordingly, an empty "content" that defines no media types is an error. Likewise a "content" that defines multiple media types. But such invalid parameters are accepted and no error is reported.

Here's how to reproduce this problem:

```java

    @Test
    public void testParamContent() {
        String json =
            "{"
            + "  \"openapi\": \"3.0.0\","
            + "  \"info\": {"
            + "    \"title\": \"Operations\","
            + "    \"version\": \"0.0.0\""
            + "  },"
            + "  \"paths\": {"
            + "    \"/operations\": {"
            + "      \"post\": {"
            + "        \"parameters\": ["
            + "          {"
            + "            \"name\": \"param0\","
            + "            \"in\": \"query\","
            + "            \"content\": {"
            + "            }"
            + "          },"
            + "          {"
            + "            \"name\": \"param1\","
            + "            \"in\": \"query\","
            + "            \"content\": {"
            + "              \"text/plain\": {"
            + "              }"
            + "            }"
            + "          },"
            + "          {"
            + "            \"name\": \"param2\","
            + "            \"in\": \"query\","
            + "            \"content\": {"
            + "              \"text/plain\": {"
            + "              },"
            + "              \"application/json\": {"
            + "                \"schema\": {"
            + "                  \"type\": \"object\""
            + "                }"
            + "              }"
            + "            }"
            + "          }"
            + "        ],"
            + "        \"responses\": {"
            + "          \"default\": {"
            + "            \"description\": \"None\""
            + "          }"
            + "        }"
            + "      }"
            + "    }"
            + "  }"
            + "}"
            ;
        OpenAPIV3Parser parser = new OpenAPIV3Parser();
        SwaggerParseResult result = parser.readContents(json, null, null);
        Operation post = result.getOpenAPI().getPaths().get( "/operations").getPost();

        Parameter param0 =
            post.getParameters().stream()
            .filter( p -> "param0".equals( p.getName()))
            .findFirst()
            .orElseThrow( () -> new IllegalStateException( "Can't find parameter=param0"));
        assertEquals( param0.getContent().size(), 0, "Empty content size"); // No! Should report an error

        Parameter param1 =
            post.getParameters().stream()
[...]

PR #1647, opened 2022-02-02
OpenAPIDeserializer: Report an error when parameter "content" definition is invalid

kerrykimbrough:
Fixes issue #1646


gracekarina:
thanks @kerrykimbrough

PR #1707, opened 2022-04-09
Fixed #1706

### LINKED ISSUE #1706: Parser is not resolving the requestBody to make it inline as its doing for responses
**MetaData** :
- Swagger Parser Version : 2.0.31
- OAS Version: 3.0.0
- Simple Reproducible Spec is pasted at the end

**Reproducible Code Snippet** :
```
io.swagger.v3.parser.core.models.ParseOptions parseOptions = new io.swagger.v3.parser.core.models.ParseOptions();
parseOptions.setResolve(true);
parseOptions.setFlatten(true);
SwaggerParseResult result = new OpenAPIParser().readLocation("<PATH TO SPEC>\\OAS3_Inline_Request_Response_Usecase.json", null, parseOptions);
```

**Context** : 
We're using the above snippet for the following reasons and explanation against it -
1. resolve is true: We have usecases where there are external json references and resolve is best fit for us to solve the problem of localizing the external references.
2. flatten is true: We have some usecases that have inline schema definitions for requestBody and responses in an operation. And we need to make them ref locally from components section. And for this flatten is the best fit.

**Issue and Observations** :
- We basically were using the spec below expecting Swagger Parser to flatten up the inline schemas for my referenced requestBody and response, and  then make them refer from components section. 
- We observed that the SwaggerParser works fine for responses in flattening up but NOT for requestBody.
- We debugged the flatten implementation, and was shocked that it only fattens the _path inline_ schemas and the schemas under _components/schema_ section, then how is it working for responses. 
- We then suspected that there's something in the _resolve_ implementation that is making responses work.
- And bingo, we found it was the exact case and [here's](https://github.com/swagger-api/swagger-parser/blob/e29a4f44ea53afe35a32c5fc8bf49012a857593f/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java#L54) the link pointing to that code. I dug deep to find the commit and here it is --> [commit](https://github.com/swagger-api/swagger-parser/commit/ab0fbe9f56f12c3399a1d71df87465344e7c612d#diff-8ea24f208841d2105efdceef4d2b24f670cff0b6d06c29dda48dd3a11125a612)
- So its not doing the same for request.
- Hey @gracekarina, I also have a fix for the same, its on the similar lines as its done for responses, I'll raise the PR soon.

**Impact and why we can't use ResolveFully** :
- This very crucial for our Q2 roadmap items.
- We are not able to resolve the inline schemas for the referenced requestBody from components section.
- We can't use resolveFully because it will make everything inline and then we intend to flatten every inline thing, which will rename the older models local references that were not inline, to the newer flattened up naming local reference. I am afraid that using resolveFully and then flatten will lead us to some regressions in our product. Making use of resolve and flatten is the best combination for us, or rather the correct one as we don't wan't to unnecessarily over process using resolveFully.
- The sample we've attached is very basic, and we have very complex usecases that have this inline flavour of referenced requestBody.
- Hey @gracekarina, Please request to help merge the similar changes I am raising in the PR for the requestBody.

**OAS3 Spec** :
```
{
	"openapi": "3.0.0",
	"info": {
		"version": "1.0",
		"title": "Resource service",
		"description": "Resource service"
	},
	"servers": [
		{
			"url": "http://localhost:8088/"
		}
	],
	"paths": {
		"/resource": {
			"post": {
				"summary": "postResource",
				"description": "postResource",
				"operationId": "post-resource",
				"parameters": [],
				"requestBody": {
					"$ref": "#/components/requestBodies/ResourceBodyPost"
				},
				"responses": {
					"200": {
						"$ref": "#/components/responses/ResourceSuccessResponse"
					}
				}
			}
		}
	},
	"components": {
		"requestBodies": {
			"ResourceBodyPost": {
				"content": {
					"application/json": {
						"schema": {
							"type": "object",
							"required": [
								"rqname"
							],
							"properties": {
								"rqname": {
									"type": "string"
								},
								"rqnumber": {
									"type": "integer",
									"format": "int32"
								},
								"rqflag": {
									"type": "boolean"
								}
							}
						}
					}
				}
			}
		},
		"responses": {
			"ResourceSuccessResponse": {
				"description": "a Element to be returned",
				"content": {
					"application/json": {
						"schema": {
							"type": "object",
							"required": [
								"rname"
							],
							"properties": {
								"rname": {
									"type": "string"
								},
								"rnumber": {
									"type": "integer",
									"format": "int32"
								},
								"rflag": {
									"type": "boolean"
								}
							}
						}
					}
				}
			}
		},
		"schemas": {}
	}
}
```

mma5997:
This makes the request body inline as we're already doing it for responses.
This will help flatten api to leverage the inline schema for this resolved request body which has an inline schema inside it.


mma5997:
Just saw that 2 tests failed. 
Hmm seems we'd need to fix those 2.
But looks everything else is smooth.

mma5997:
updated the test files as per the primary changes done in this PR.

mma5997:
Hi @gracekarina 
I am done with all the changes, please can you review and help proceeding with the workflows ahead.
Thanks for your time :)

gracekarina:
@mma5997 Hi!, sorry it took so long for me to reply, I was reviewing the changes and they seem ok, but we should not change a functionality that could break so many dependent users of parser's current behavior, so, what we can do is add an option in ParseOptions like flatten, resolveFully, etc,  and if anyone needs it, configure parser with the option set to true, and that's it. https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-core/src/main/java/io/swagger/v3/parser/core/models/ParseOptions.java#L6 if you need any help adding the option let me know. Thanks for this PR!!!

mma5997:
No issues, I was eagerly waiting for your response :) 
Thanks for the suggestion @gracekarina  , you're right, change in behaviour would impact the backward compatibility and the existing user-base. 

So, I've tried adding a new parseOption to toggle the resolution of the requestBody accordingly.
Please can you help reviewing it. Please do let me know your thoughts.

PS:
Maybe later I'll need to add separate unit test for this scenario. 
Also I had made a similar parseOption for response resolution (which happens by default so its default value is true) which might be on similar lines to requestBody. But seems there's a test - `io.swagger.v3.parser.processors.OperationProcessorTest.testProcessOperation(List<Parameter>, List<Parameter>, ApiResponse, ApiResponse)` failing due to some reasons, and I thus made the changes only for requestBody apparently.

gracekarina:
Hi, @mma5997  thanks for this approach, at the moment I have to do a thorough review of this solution, as it seems, it's still putting inline the requestBodies, and the option is not being set in the test. I will work on a different branch and commit the changes, they will be in the next release.

mma5997:
Thanks @gracekarina 😃 , sure!
just to clarify and summarize once about the final changes [here](https://github.com/swagger-api/swagger-parser/pull/1707/files)

1. I have just added a member var `resolveRequestBody` in ParseOptions class, whose default value is `false` because its the normal default behaviour that, the parser doesn't put the request body inline. So the existing parser user-base will not face any problem. The setter for that var (`resolveRequestBody`) will help setting the toggle for putting the requestBody inline which solves #1706.
2. In OperationProcessor, I've added the check to see if the `resolveRequestBody` is set to `true` using the getter for the same and then put requestBody inline IFF its set to true.
3. Added the getter in ResolverCache class for getting the ParseOptions instance which will help us check if `resolveRequestBody` is set to `true` or not in the above point 2.

The `resolveRequestBody` ParseOption is not needed to be set explicitly in the test or anywhere else because the default behaviour is kept intact by defaulting  `resolveRequestBody` to `false` as aforementioned.
I haven't added any unit test for `resolveRequestBody` set to true, yet.

Thanks again 😄 !

gracekarina:
Hi can you please add the test for the option set to true? thanks

mma5997:
Hello @gracekarina,
Thanks for your time, I've added a simple test, please review :)

modules/swagger-parser-core/src/main/java/io/swagger/v3/parser/core/models/ParseOptions.java
```diff
@@ -36,6 +37,18 @@ public boolean isResolveFully() {
...
+		return resolveRequestBody;
+	}
+    
+    /**
+     * If set to true, will help resolving the requestBody as inline, provided resolve is also set to true.
+     * Default is false because of the existing behaviour.
+     */
+	public void setResolveRequestBody(boolean resolveRequestBody) {
+		this.resolveRequestBody = resolveRequestBody;
+	}
```
frantuma, line 51: Would need to use spaces instead of tabs and correct indentation

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
```diff
@@ -68,11 +68,10 @@ public class ResolverCache {
     private Map<String, String> externalFileCache = new HashMap<>();
     private List<String> referencedModelKeys = new ArrayList<>();
     private Set<String> resolveValidationMessages;
-
     private final ParseOptions parseOptions;
 
 
-    /*
+	/*
```
frantuma, line 74: Would need to use spaces instead of tabs and correct indentation

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
```diff
@@ -394,4 +393,8 @@ public Map<String, String> getExternalFileCache() {
     public Map<String, String> getRenameCache() {
         return Collections.unmodifiableMap(renameCache);
     }
+    
+    public ParseOptions getParseOptions() {
+		return parseOptions;
+	}
```
frantuma, line 399: Would need to use spaces instead of tabs and correct indentation

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java
```diff
@@ -40,8 +39,19 @@ public void processOperation(Operation operation) {
...
+        	if (requestBody.get$ref() != null && cache != null && cache.getParseOptions() != null && cache.getParseOptions().isResolveRequestBody()) {
+        		requestBodyProcessor.processRequestBody(requestBody);
+        		RefFormat refFormat = computeRefFormat(requestBody.get$ref());
+        		RequestBody resolvedRequestBody = cache.loadRef(requestBody.get$ref(), refFormat, RequestBody.class);
+
+                if (resolvedRequestBody != null) {
+                	requestBody = resolvedRequestBody;
+                	operation.setRequestBody(resolvedRequestBody);
+                }
+        	}
```
frantuma, line 54: Would need to use spaces instead of tabs and correct indentation

modules/swagger-parser-v3/src/test/resources/issue-1706/SimpleRequestResponseRef.json
```diff
@@ -0,0 +1,87 @@
+{
```
frantuma, line 1: Would need to use spaces instead of tabs

frantuma (commented):
Hi @mma5997 thanks a lot! can you possibly just fix the indent/tabs things mentioned in review comments?

modules/swagger-parser-v3/src/test/resources/issue-1706/SimpleRequestResponseRef.json
mma5997: done

modules/swagger-parser-core/src/main/java/io/swagger/v3/parser/core/models/ParseOptions.java
mma5997: Done!

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
mma5997: Done!

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
mma5997: Done!

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java
mma5997: Done!

modules/swagger-parser-v3/src/test/resources/issue-1706/SimpleRequestResponseRef.json
mma5997: Done!

mma5997:
Thanks @frantuma, 
please can you review once and let me know if anything's needed, I'll be more than happy :)

modules/swagger-parser-v3/src/test/resources/issue-1706/SimpleRequestResponseRef.json
frantuma: @mma5997 it seems that json file still has tabs instead of spaces. Sorry about that, it's to get PRs as clean as we can

mma5997: Thanks for pointing it out @frantuma. 
Sorry, my bad, Eclipse formatting likes tabs it seems XD. No issues at all, I agree with you, it should be clean definitely. I have fixed it.

frantuma:
Thanks for your contribution, merged!

Issue #1205, opened 2019-09-25
Enum string value "YES" becomes "true"


piebur:
I have encountered an exception which looks suspicious while utilizing the String enum property with the value YES. The result is that YES becomes true somehow?

My guess is that is has to do with the conversion from YAML to JSON.

Reproduce:
1. Unzip the attached file
2. Copy the new-issue.yaml to the following location: swagger-parser\modules\swagger-parser-v3\src\test\resources
3. Replace OpenAPIV3ParserTest.java in the following location: swagger-parser\modules\swagger-parser-v3\src\test\java\io\swagger\v3\parser\test
4. Execute the test case testNewIssue(), the result is failure.

A work around could be to replace YES with some other string else or just add " to make it a string

Reference: https://github.com/vert-x3/vertx-web/issues/1394

[reproduce.zip](https://github.com/swagger-api/swagger-parser/files/3651917/reproduce.zip)

webron:
Yup, that's possible behavior when using YAML 1.1. `YES` can be considered as `true` (also `ON` and some other things). We should disable this kind of parsing though, and I was sure we did before. Perhaps it got 'lost' in the new version of the parser. Which version do you use?

As mentioned in the linked ticket, for now, the workaround is to explicitly say it's a string to avoid the parsing to consider it a boolean.

piebur:
I am not sure what YAML version OpenApi uses, but I have within the new-issues.yaml to use OpenApi 3.0. Looking at the code OpenAPIV3Parser.java converts the YAML using io.swagger.v3.parser.OpenAPIV3Parser.YAML_MAPPER.

webron:
That's actually not true, @piebur. JSON Schema doesn't work that way. The `type` has no impact on the `enum` or vice versa. You can define the type to be integer, and then have the enum contain only string values. The definition is legal and only means nothing will validate against it (except for no value, if that's allowed). An enum can also have multiple types in it. In short, we can't infer the type of the value in the enum from the defined type.

manojkva:
I could fix it using the On/Off enum (the ones with issue) value to be under a double quote ("). Example
enum:
- "On"
- "Off"

hkosova:
Related: https://github.com/swagger-api/swagger-core/issues/3372

davidmoten:
still happening in swagger-parser 2.1.16. Can this be fixed please?

davidmoten:
[...]

PR #1709, opened 2022-04-14
Converting yes no on off enum values without quotes to booleans problem

MateuszPol:
Hi, we are getting problems while using `DesrializationUtils`. When using `DeserializationUtils.readYamlTree(...)` method, YAML definition that contains not quoted enum values like Yes/No/On/Off are converted to boolean values, which seems not correct behaviour.

I wrote a UnitTest to with example of YAML file and expected result.


gracekarina:
related https://github.com/swagger-api/swagger-parser/issues/1205

Issue #1711, opened 2022-04-21
Inconsistent content encodings allowed


kerrykimbrough:
From the OpenAPI spec for [`MediaType`](https://spec.openapis.org/oas/v3.0.1#media-type-object):

> encoding: A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property.

It is an error if the `encoding` map contains a key that is not defined as a property name in the content (`object`) schema. But this error is not recognized. The following test fails because it doesn't report the encoding for non-existent property "Z" as an error.

```
@Test
public void testEncodingNotApplicable() {
    String json =
        "{"
        + "  \"openapi\": \"3.0.0\","
        + "  \"info\": {"
        + "    \"title\": \"Encodings\","
        + "    \"version\": \"0.0.0\""
        + "  },"
        + "  \"paths\": {"
        + "    \"/resource\": {"
        + "      \"post\": {"
        + "        \"requestBody\": {"
        + "          \"content\": {"
        + "            \"multipart/form-data\": {"
        + "              \"schema\": {"
        + "                \"type\": \"object\","
        + "                \"properties\": {"
        + "                  \"X\": {"
        + "                    \"type\": \"integer\""
        + "                  },"
        + "                  \"Y\": {"
        + "                    \"type\": \"integer\""
        + "                  }"
        + "                }"
        + "              },"
        + "              \"encoding\": {"
        + "                \"Z\": {"
        + "                  \"contentType\": \"application/json\""
        + "                }"
        + "              }"
        + "            }"
        + "          }"
        + "        },"
        + "        \"responses\": {"
        + "          \"default\": {"
        + "            \"description\": \"None\""
        + "          }"
        + "        }"
        + "      }"
        + "    }"
        + "  }"
        + "}"
        ;
    OpenAPIV3Parser parser = new OpenAPIV3Parser();
    SwaggerParseResult result = parser.readContents(json, null, null);
    assertEquals( result.getMessages().size(), 1, "Error messages");
}
```

frantuma:
fixed in #1712

PR #1712, opened 2022-04-21
getMediaType: Report an error for any encoding property not defined in the schema.

kerrykimbrough:
Fixes issue #1711


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -1357,8 +1357,12 @@ public MediaType getMediaType(ObjectNode contentNode, String location, ParseResu
 
 		ObjectNode encodingObject = getObject("encoding", contentNode, false, location, result);
 		if (encodingObject != null) {
-			mediaType.setEncoding(getEncodingMap(encodingObject, String.format("%s.%s", location, "encoding"),
-					result));
+            String encodingLocation = String.format("%s.%s", location, "encoding");
+			mediaType.setEncoding(getEncodingMap(encodingObject, encodingLocation, result));
+
+            mediaType.getEncoding().keySet().stream()
+                .filter( ep -> Optional.ofNullable( mediaType.getSchema().getProperties()).map( p -> !p.containsKey( ep)).orElse( true))
```
frantuma, line 1364: this would break for a null schema (schema not defined in media type)

frantuma, line 1364: This should be only executed for inline schemas, by skipping it when a schema `$ref` field is existing and non empty, as properties in this case can be defined in the referenced schema

frantuma (commented):
@kerrykimbrough Thanks for reporting and working on this. A couple of inline comments before being able to merge it

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
kerrykimbrough: Thanks, good catch! I've revised the code and test accordingly.

frantuma:
Thanks!

Issue #1169, opened 2019-08-13
Using multiple path params in a same URI segment of a path in swagger gives unexpected validation message.


simpijain:
Using multiple path params in a same URI segment of a path in swagger fails with a validation message - 

[attribute paths.'/test/{id}-{name}'. Declared path parameter id}-{name needs to be defined as a path parameter in path or operation level]

In the older version of swagger parser jar this validation was not there but now we are getting with latest version of swagger parser jar 1.0.45. 
'/test/{id}-{name}' is a correct path for REST, 'id' and 'name' should be considered as 2 different path parameters not one. Please look into this,

Attaching the swagger below.  

[swagger7.txt](https://github.com/swagger-api/swagger-parser/files/3495862/swagger7.txt)

DavidShiel:
Using the latest version of the OAS3 parser (swagger-parser-v3-2.0.13.jar), just found the same issue with this API - https://api.apis.guru/v2/specs/gov.bc.ca/bcgnws/3.x.x/openapi.json

Issue with the following path: **/names/{nameId}.{outputFormat}**
Reported error: "attribute paths.'/names/{nameId}.{outputFormat}'. Declared path parameter nameId}.{outputFormat needs to be defined as a path parameter in path or operation level"

io.swagger.v3.parser.util.OpenAPIDeserializer.getPaths is incorrectly assuming that all path parameters are separated by '/'

gracekarina:
Thanks @simpijain and @DavidShiel  for reporting this issue, is now fixed. Please let us know if this keeps happening.

michaelkourlas:
I think this issue should be re-opened -- I don't think just checking for certain characters that happen to be used as separators (e.g. "." and "-") is a reliable solution. For example, my spec uses "," as a separator, so it gets caught by this.

webron:
@michaelkourlas can you share an example?

michaelkourlas:
Well, a path of the form `/test/{id},{name}`, instead of `/test/{id}-{name}` or `/test/{id}.{name}`, will cause the same failure. But `,` is not the only possible separator -- any character or sequence of characters could be used.

The core problem is that the code assumes that a path can be split into "parts", with exactly one path parameter per "part":

```java
String[] eachPart = pathName.split("[-/.]+");
```

But this just isn't true. It should just look for path parameters (`{`, then parameter name, then `}`) throughout the whole path without doing any string splitting.

PR #1718, opened 2022-05-04
Bug/issue1169 nosplit v3

exhuma:
This contains a fix for path-element fixing (Delegated PR for @lucian-r)


exhuma:
Related to #1169

PR #1719, opened 2022-05-04
fix for issue 1169 that works without explicit separators

exhuma:
This contains a fix for path-element fixing (Delegated PR for @lucian-r)

fix for v3 is in #1718


exhuma:
Related to #1169

PR #1733, opened 2022-05-24
Paths defined as $ref wont add parameters to each operation

frankbille:
## Problem

When organizing path definitions in different files like the example below, the parameters will not be passed from the top-level `/ref/test/{id}/toplevelparam` definition to the `GET` operation.

`api.yaml`:
```yaml
openapi: 3.0.0

paths:
  /ref/test/{id}/toplevelparam:
    $ref: "./test-endpoints.yaml#/paths/~1ref~1test~1{id}~1toplevelparam"
```

`test-endpoints.yaml`:
```yaml
openapi: 3.0.0

paths:
  /ref/test/{id}/toplevelparam:
    parameters:
      - in: path
        name: id
        required: true
        schema:
          type: string
          enum:
            - one
            - two
    get:
      summary: Test of path params defined on top level
      responses:
        200:
          description: OK
      tags:
        - Tests
```

The issue is that the passing of parameters to the operations is done before parsing of the `$ref`s.

## Solution

By calling `addParametersToEachOperation` after all the `$ref`s has been processed, the parameters can be correctly added.

I have included simple tests of `PathProcessor` for this specific case. I am not sure if I have structured the tests correctly or if they make more sense to include in `OpenAPIResolverTest`


frankbille:
There are two tests that fail as a consequence of the change to `PathProcessor`, because they expect there to be parameters on top level. When [PathProcessor.addParametersToEachOperation](https://github.com/swagger-api/swagger-parser/blob/3f89a9a624394537098d07b6a1c1f676e747ad9e/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/PathsProcessor.java#L93) is called it ends by setting the top level parameters to null.

I think the test needs to be changed because it would never have passed if the schema was all internal.

gracekarina:
thanks for this PR!

Issue #1758, opened 2022-06-22
No validation errors for when a header references a schema in components


ponelat:
The following definition is invalid..
```yml
openapi: 3.0.3

info: 
  title: Missing validation rule for schemas in Headers.
  version: 1.0.0
  
paths: 
  /foo:
    get:
      description: ok
      parameters:         
      - $ref: '#/components/schemas/xFoo'
      responses:
        default:
          description: ok
          headers:
            three:
              $ref: '#/components/schemas/xFoo'
           
components:
  schemas:
    xFoo:
      type: string
      description: This isn't validated correctly

```

In it the parameter (and header) are both referencing `#/components/schemas/xFoo`, and `xFoo` isn't a valid header/paramter. A valid would header/parameter needs the `schema` property, e.g.:
```yml
xFoo:
  schema:
    type: string
    description: This is a valid header
```
Tested with https://validator.swagger.io (in case there is a discrepency).

Is there a decision to not validate the fully resolved definition? Or a decision to validate against improper paths (i.e.,  referencing a JSON Pointer that looks fishy, even if technically it's allowed by the spec to reference any path).

cc @frantuma @gracekarina

gracekarina:
@ponelat This issue has been addressed in https://github.com/swagger-api/swagger-parser/pull/1760, please let us know if this keeps happening.

PR #1760, opened 2022-06-25
fix and test for issue 1758

gracekarina:
Fix and test for issue #1758.

Adds a validation message for parameters and headers when ref field in **internal references** are not being pointed at the right components tag.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -1905,10 +1905,13 @@ public class OpenAPIDeserializer {
 				parameter = new Parameter();
 				String mungedRef = mungedRef(ref.textValue());
 				if (mungedRef != null) {
-					parameter.set$ref(mungedRef);
-				} else {
-					parameter.set$ref(ref.textValue());
-				}
+                    parameter.set$ref(mungedRef);
+                }else {
+                    parameter.set$ref(ref.textValue());
+                }
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/parameters")||ref.textValue().startsWith("#/components/headers"))) {
+                    result.warning(location, "$ref is not pointing to #/components/parameters");
+                }
 				if (result.isOpenapi31()) {
 					String desc = getString("description", obj, false, location, result);
 					if (StringUtils.isNotBlank(desc)) {
@@ -2115,6 +2118,10 @@ public class OpenAPIDeserializer {
 				} else {
 					header.set$ref(ref.textValue());
 				}
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/parameters")||ref.textValue().startsWith("#/components/headers"))) {
+                    result.warning(location, "$ref is not pointing to #/components/headers");
+                }
+
 				if (result.isOpenapi31()) {
 					String desc = getString("description", headerNode, false, location, result);
 					if (StringUtils.isNotBlank(desc)) {
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -958,6 +958,9 @@ public class OpenAPIDeserializer {
 				} else {
 					pathItem.set$ref(ref.textValue());
 				}
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/pathItems"))) {
+                    result.warning(location, "$ref target "+ref.textValue() +" is not of expected type");
+                }
 				if(result.isOpenapi31()){
 					String value = getString("summary", obj, false, location, result);
 					if (StringUtils.isNotBlank(value)) {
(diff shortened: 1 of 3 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -1905,10 +1905,13 @@ public Parameter getParameter(ObjectNode obj, String location, ParseResult resul
...
 				if (mungedRef != null) {
-					parameter.set$ref(mungedRef);
-				} else {
-					parameter.set$ref(ref.textValue());
-				}
+                    parameter.set$ref(mungedRef);
+                }else {
+                    parameter.set$ref(ref.textValue());
+                }
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/parameters")||ref.textValue().startsWith("#/components/headers"))) {
```
frantuma, line 1912: Why is it accepting a ref to `#/components/headers` for parameters? (and the other way around below?)

frantuma (changes requested):

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -1905,10 +1905,13 @@ public Parameter getParameter(ObjectNode obj, String location, ParseResult resul
...
+                    result.warning(location, "$ref is not pointing to #/components/parameters");
```
frantuma, line 1913: possibly change to something like `$ref target #/components/headers/whatever is not of expected type 'header'`. @ponelat possibly has some even better wording

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
gracekarina: headers and parameters do share the same structure, and they can use with one or the other.
Docs: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#header-object
tests that use it: https://github.com/swagger-api/swagger-parser/blob/707adcb2fe3a2d31e8c915d9e4cacdb3b176f7c2/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java#L524

gracekarina:
@ponelat @frantuma I also did the validation in deserializer of examples, and securitySchemes, but not 100% sure it applies here.
If you have a sample of a ref example being use, and securitySchemes using the ref `#/components/...` please let me know, I couldn't find any.
Thanks

ponelat:
@gracekarina `securitySchemas` aren't referenced using `$ref`, instead they're referenced via the `security: []` fields in operations and globally.

Here is an example of using `$ref` to point to an example. Not that they're not for schemas, but for response/requestBodies, i.e., `examples`, see: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#exampleObject

```yaml
openapi: 3.0.3
info:
  title: $ref of an "Example".
  version: 1.0.0
paths:
  /foo:
    get:
      description: ok
      responses:
        '200':
          description: ok
          content:
            application/json:
              schema:
                type: object
              examples:
                one:
                  $ref: '#/components/examples/Ex'
              
components:
  examples:
    Ex:
      summary: The value one.
      value: 
        one: '1'
  ```

File: modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java [hunk 3]
```diff
@@ -958,6 +958,9 @@ public PathItem getPathItem(ObjectNode obj, String location, ParseResult result)
 				} else {
 					pathItem.set$ref(ref.textValue());
 				}
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/pathItems"))) {
+                    result.warning(location, "$ref target "+ref.textValue() +" is not of expected type");
```
frantuma, line 962: I would detail what is the expected type, like `"$ref target "+ref.textValue() +" is not of expected type pathItem");`. Same applies to similar code below

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -1905,10 +1914,13 @@ public Parameter getParameter(ObjectNode obj, String location, ParseResult resul
...
 				if (mungedRef != null) {
-					parameter.set$ref(mungedRef);
-				} else {
-					parameter.set$ref(ref.textValue());
-				}
+                    parameter.set$ref(mungedRef);
+                }else {
+                    parameter.set$ref(ref.textValue());
+                }
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/parameters")||ref.textValue().startsWith("#/components/headers"))) {
```
frantuma, line 1921: cdoe style, would replace `meters")||ref.textVal` with `meters") || ref.textVal`

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -2115,6 +2127,10 @@ public Header getHeader(ObjectNode headerNode, String location, ParseResult resu
 				} else {
 					header.set$ref(ref.textValue());
 				}
+                if(ref.textValue().startsWith("#/components") && !(ref.textValue().startsWith("#/components/parameters")||ref.textValue().startsWith("#/components/headers"))) {
```
frantuma, line 2130: cdoe style, would replace `meters")||ref.textVal` with `meters") || ref.textVal`

frantuma (commented):
@gracekarina couple of minor comments, when addressed please merge

Issue #1757, opened 2022-06-22
Component schema with a value of true or false throws an error


spacether:
Component schema with a value of true or false throws and error

```
openapi: 3.0.3
info:
  title: openapi 3.0.3 sample spec
  description: sample spec for testing openapi functionality, built from json schema
    tests for draft6
  version: 0.0.1
paths: {}
components:
  schemas:
    BooleanSchemaTrue: true
    BooleanSchemaFalse: false
```
Throws the error:
BooleanSchemaTrue is not of type `object`
BooleanSchemaFalse is not of type `object`

For openapi v3.0.3 this is probably correct, I got these results using swagger-parser 2.0.31
But for v3.1.0 specs, should we get this result?

frantuma:
thanks for reporting this. Fixed in https://github.com/swagger-api/swagger-parser/pull/1768 for 3.1 specs

PR #1768, opened 2022-07-05
refs #1757 - add boolean schema support and fix type inference

frantuma:


modules/swagger-parser-v3/src/test/resources/issue-1761.yaml
```diff
@@ -0,0 +1,214 @@
...
+    EnumWith0DoesNotMatchFalse:
+      enum:
+        - 0
+    EnumWith1DoesNotMatchTrue:
+      enum:
+        - 1
+    NulCharactersInStrings:
+      enum:
+        - "hello\0there"
+  x-schema-test-examples:
```
spacether, line 58: You probably don't need this vendor extension test data because your project is not using it

PR #1795, opened 2022-09-13
Issue 1454

bfreuden:
It fixes only one of the failing cases of the provided tests though.
I didn't consider cases where the default value is defined on the enum and expected to be inherited by the property.


gracekarina:
thanks for your contribution

davidmoten:
Would project owners please edit pr title so that is has some meaning before it ends up in the release notes? Even the pr description is completely meaningless.

PR #1819, opened 2022-09-29
Update ReferenceUtils to better support windows filesystem

snaggedagge:
Occurs on windows filesystems:

when absoluteUri is something like "C:/openapi.yml", the "scheme" part in URI object is resolved into "C". So it passes through all the statements and lands on an exception being thrown. A more resilient approach is to always default to reading from file when the scheme cannot be determined


gracekarina:
Thanks!

Issue #1802, opened 2022-09-20
ResolveFully ignores properties in schema with combinators


devdevx:
## Parser version: 
2.1.2

## Spec:

```
openapi: "3.0.0"
info:
  version: 1.0.0
  title: 0001_allOffProps

paths:
  /sample_resource:
    post:
      operationId: createResOne
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/create_res_one"
      responses:
        '201':
          description: Ok
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/standard_response_res_one"
components:
  schemas:
    standard_response_result:
      properties:
        result:
          type: object
          properties:
            status:
              type: boolean
            http_code:
              type: integer
            errors:
              type: array
              items:
                $ref: '#/components/schemas/standard_error'
            info:
              type: string
            trace_id:
              type: string
            num_elements:
              type: integer
          required:
            - status
            - http_code
            - trace_id
    standard_error:
      type: object
      properties:
        code:
          type: integer
        message:
          type: string
    standard_response_res_one:
      type: object
      allOf:
        - $ref: '#/components/schemas/standard_response_result'
      properties:
        data:
          properties:
            name:
              type: string
    create_res_one:
      type: object
      properties:
        name:
          type: string
```

## Code:

```
        ParseOptions parseOptions = new ParseOptions();
        parseOptions.setResolveFully(true);
        OpenAPI openAPI = new OpenAPIV3Parser().read("api.yaml" + fileName, null, parseOptions);
```

##  Expected result:
  
  The schema `standard_response_res_one` has 2 properties: `data` and `result`
  
##  Current result:
  
  The schema `standard_response_res_one` only has the `result` property
  
##  Investigation about the issue cause:
  
After debugging the issue seems that in the class `ResolverFully` we need to add the properties of the composedSchema to the result schema (https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java#L405)

Issue #1777, opened 2022-08-03
'allOf' not processed correctly for property


devdevx:
Example with one schema property:

```
openapi: "3.0.0"
info:
  title: API-Template
  description: |
    Single or multilined API description. Can be written on HTML or [CommonMark](http://commonmark.org/help/)
  version: v1

tags:
  - name: Customers
    description: "Operations and resources related to customers"

paths:
  /customers/{customerId}:
    parameters:
      - $ref: '#/components/parameters/customerIdPathParam'
    get:
      summary: Retrieve customer information
      description: Description for operation that allows retrieve customer information
      operationId: retrieveCustomerInfo
      tags:
        - Customers
      responses:
        '200':
          description: Customer response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/customer'

components:
  parameters:
    customerIdPathParam:
      name: customerId
      in: path
      required: true
      description: The id of the customer
      schema:
        $ref: '#/components/schemas/uuid'

  schemas:
    uuid:
      type: string
      format: uuid
      minLength: 36
      maxLength: 36
      pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'

    customer:
      type: object
      properties:
        id:
          description: The id of the customer
          allOf:
           - $ref: '#/components/schemas/uuid'
          x-apigen-mapping:
            field: id
```

Code executed:

´´´
import io.swagger.v3.core.util.Yaml;
import io.swagger.v3.oas.models.OpenAPI;
import io.swagger.v3.parser.OpenAPIV3Parser;
import io.swagger.v3.parser.core.models.ParseOptions;
import org.apache.commons.io.IOUtils;

import java.io.IOException;
import java.io.InputStream;
import java.nio.charset.StandardCharsets;

public class Main {
    public static void main(String[] args) throws IOException {
        String api;
        try (InputStream is = Main.class.getClassLoader().getResourceAsStream("template_error.yaml")) {
            api = IOUtils.toString(is, StandardCharsets.UTF_8);
        }
        OpenAPIV3Parser parser = new OpenAPIV3Parser();
        ParseOptions options = new ParseOptions();
        options.setResolve(true);
        options.setResolveCombinators(true);
        options.setResolveFully(true);
        options.setFlatten(true);
        OpenAPI openAPI = parser.readContents(api, null, options).getOpenAPI();
[...]

PR #1823, opened 2022-10-10
Extend resolve fully and solve issues #1777 and #1802

devdevx:
@gracekarina This PR means to fix some missing information in the resolve fully functionality. We have them fixed together but let me know if you prefer the changes in 2 separated PR.


devdevx:
My changes breaks a test (testIssue1266) that didn't expect the `type: object` explicit (now is implicit by definition). Should I update my changes to avoid set the type when is object or can I update the expected result of the failing test? If I am not wrong `type: object` and no explicit type means the same.

devdevx:
@gracekarina just let me know if you need some changes in the PR

gracekarina:
Hi @devdevx sorry it took so long. Thanks for this PR!

PR #1890, opened 2023-03-07
url and identifier in the license field are handled incorrectly

bweben:
Currently, swagger-parser will error with the following message `attribute info.license.identifier is missing` when parsing an OpenApi 3.1 document where `license.url` and `license.name` is set (see the example document below).
According to the official OpenApi 3.1 schema it must have either the `license.url` or the `license.identifier` set but not both.
This PR will fix swagger-parser to work according to the schema.

Example OpenApi 3.1 which currently fails:
```yaml
openapi: 3.1.0
servers:
  - url: 'http://abc:5555/mypath'
info:
  version: '1.0'
  title: dd
  license:
    name: test
    url: http://example.com
paths:
  /resource1/Id:
    post:
      description: ''
      operationId: postOp
      responses:
        '200':
          description: Successful
        '401':
          description: Access Denied
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/mydefinition'
        required: true
components:
  schemas:
    mydefinition: {}
```

Link to the OpenApi 3.1 schema:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#licenseObject

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -1305,9 +1305,17 @@ public class OpenAPIDeserializer {
 		}
 
 		if (result.isOpenapi31()) {
-			value = getString("identifier", node, true, location, result);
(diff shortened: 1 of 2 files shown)


frantuma:
Thanks!

Issue #1891, opened 2023-03-13
Reference resolution fails for project with mix of subfolders and classpath references


douglasbgray:
if a project is configured as follows ...

```text
src/main/resources/
    ├── openapi.yaml
    ├── types/
        ├── local-types.yaml
```

... and a model in `openapi.yaml` has a reference to a model in `local-types.yaml`, and in turn, that model is an alias to a model found in a classpath resource, then the reference is not resolved. 

### Example

**openapi.yaml** (root contract)
```yaml
LocalModel:
  type: object
  properties:
    sharedModelField:
      $ref: 'types/local-types.yaml#/components/schemas/TypesModel'
```

**local-types.yaml** (exists in `types` folder sibling to root contract)
```yaml
TypesModel:
  $ref: '/shared-types.yaml#/components/schemas/SharedModel'
```

**shared-types.yaml** (this is in shared-types.yaml resource in the classpath, not in file system)
```yaml
SharedModel:
  type: object
  properties:
    name:
      type: string
```

The parser attempts to find resolve the reference to `/shared-types.yaml` by prepending the path of the local types directory, so it looks for `types/shared-types.yaml` and does not find it. This leave the reference unresolved.

Since the parser is unable to resolve the references, this affects downstream code generation, which creates invalid types for some properties.

A project with a unit test demonstrating the issue is here - https://github.com/douglasbgray/cp-example

This is first observed in version 2.0.24 and still exists in the most recent version 2.1.12. 

Here is a PR with a fix and corresponding test - https://github.com/swagger-api/swagger-parser/pull/1892

douglasbgray:
PR was merged, issue was resolved. Closing.

PR #1892, opened 2023-03-13
Issue-1891: Fix for mixed folder and classpath references not resolving.

douglasbgray:
This is a fix for #1891 

Normally, in order to properly test resolving a reference with an absolute path, I would need access to a jar that can be read by the class loader. However, by using folders, I was able to simulate this by placing the file that is referred to in the contract _outside of the directory in which the rest of the test resources are located_. This causes the `ClasspathHelper` to be used to resolve the absolute reference and the test passes as desired. 

Also verified that if I remove the change to `ExternalRefProcessor` that the test fails, so I know the branch in the updated code is being executed by the test.


gracekarina (approved):
thanks so much for this PR!.

douglasbgray:
Thanks @gracekarina for accepting and merging.

Issue #1572, opened 2021-06-07
StringIndexOutOfBoundsException due to (servers - url /)


mhammadkassem:
Getting this error when parsing a file:
java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 25

	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
	at java.base/java.lang.String.substring(String.java:1874)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.getServer(OpenAPIDeserializer.java:481)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.getServersList(OpenAPIDeserializer.java:435)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.parseRoot(OpenAPIDeserializer.java:192)
	at io.swagger.v3.parser.util.OpenAPIDeserializer.deserialize(OpenAPIDeserializer.java:149)
	at io.swagger.v3.parser.OpenAPIV3Parser.parseJsonNode(OpenAPIV3Parser.java:134)
	at io.swagger.v3.parser.OpenAPIV3Parser.readContents(OpenAPIV3Parser.java:152)
	at io.swagger.v3.parser.OpenAPIV3Parser.readLocation(OpenAPIV3Parser.java:89)
	at io.swagger.parser.OpenAPIParser.readLocation(OpenAPIParser.java:16)

after debugging I noticed that the parser is failing to read 
servers:
  - url: /just/an/example
 
but when removing the dash before the url, the parsing goes fine, so he servers part becomes:
servers:
   url: /v1/analytics/time-series

jhannes:
This seems related to #1553. I tried to provoke this error when I created the PR for this. What is the context? How is OpenAPIParser.readLocation called? In the #1565 I translated paths going into OpenAPIParser from using Windows `\` path separators to `/`, which fixed the problem in my context.

mhammadkassem:
I am calling OpenAPIParser in this context:
```
ParseOptions parseOptions = new ParseOptions();
parseOptions.setResolve(true);
parseOptions.setResolveFully(true);
SwaggerParseResult result = new OpenAPIParser().readLocation(fileLocation,null,parseOptions);
OpenAPI openAPI = result.getOpenAPI();
```

note that all paths are being parsed and read correctly, the only issue here is in the servers. 
Can you please explain how the path separator is related to this issue?

jhannes:
[...]

PR #1893, opened 2023-03-14
Add test to expose issue 1572 - spaces in path to doc using relative server

ind1go:
#1572 shows that when a document is read from a location with spaces (or other URI-invalid characters) and has a relative server path, an error parsing the document occurs.

This PR starts by exposing the issue through a new test.


ind1go:
✅ Test failure as expected.

ind1go:
✅ Fix added to check that the server URL has both `{` and `}` in the case where the document path is not a valid URI.

Issue #1889, opened 2023-03-01
reference resolution fails in openapi-diff for schema reference for array items


pmhofman:
Reference resolution fails. Seems to be caused by using the same name in both openapi.yaml components/schemas and the filename of the referenced schema file.

Please see details and example in https://github.com/OpenAPITools/openapi-diff/issues/464

PR #1896, opened 2023-03-15
#1889 - Add reference to parent element

craftsman228:
Fix for the reference problem which always is null for the array and parent element https://github.com/swagger-api/swagger-parser/issues/1889.
Inside of the tickets fixed:
 - Explicit set up for the parent schema;
 - Fix for the schema detection inside the properties;


modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java
```diff
@@ -2296,10 +2291,6 @@ private OpenAPI doRelativeFileTest(String location) {
         assertEquals(refInDefinitions.getDescription(), "The example model");
         expectedPropertiesInModel(refInDefinitions, "foo", "bar");
```
craftsman228, line 2293: It's deleted cause it's not correct anymore. We shouldn't get array where we parsed schema with object.

modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java
```diff
@@ -3398,25 +3386,25 @@ public void testValidateExternalRefsTrue() {
...
-        OpenAPI openAPI = result.getOpenAPI();
-        Schema<?> localModel = openAPI.getComponents().getSchemas().get("LocalModel");
-        Schema<?> typesModel = openAPI.getComponents().getSchemas().get("TypesModel");
-        Schema<?> sharedModel = openAPI.getComponents().getSchemas().get("SharedModel");
-
-        assertEquals("#/components/schemas/TypesModel", localModel.getProperties().get("sharedModelField").get$ref());
-        assertEquals("#/components/schemas/SharedModel", typesModel.get$ref());
-        assertNotNull(sharedModel);
-    }
+//    @Test(description = "test that a model in a folder that has a ref to a model in the classpath is properly resolved.")
```
craftsman228, line 3389: This is the newest commit. Commented for now in order to save time. Will uncomment a little bit later when will not have a time.

frantuma:
@gracekarina PR reverted as it breaks in several scenarios (e.g. codegen) and IIANM also [added test](https://github.com/swagger-api/swagger-parser/pull/1896/files#diff-994c779aea4af722d351e883094df1d65116a95d2b4f4745ee4f7986cff45730R470) expects (wrongly) a ref where an array is supposed to be

PR #1911, opened 2023-04-14
SWG-7516 utilizing safeURLResolver in swagger-parser-v3

MiloszTarka:
This PR makes swagger-parser-v3 to utilizes safeUrlResolver when resolving OpenAPI 3.1 definitions.

The reason behind this change is described deeply under the following PR:
https://github.com/swagger-api/swagger-parser/pull/1910

The $refs are checked in ReferenceVisitor class, in readHTTP method.


frantuma (approved):

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
```diff
@@ -251,6 +256,8 @@ private SwaggerParseResult resolve(SwaggerParseResult result, List<Authorization
                     }
                 }
             }
+        }  catch (HostDeniedException error) {
```
CalemRoelofsSB, line 259: Do we need to rethrow this in this case or is it okay to just let is be added to the result messages in the `catch (Exception e)` block beneath?

gracekarina (approved):
LGTM!

CalemRoelofsSB (approved):
Awesome job 👍

PR #1935, opened 2023-05-24
parser-v2: Extend DeserializationUtils methods to enable providing a custom YAMLMapper

CalemRoelofsSB:
This PR adds the possibility to set a custom YAML mapper for parsing OAS 2.0 documents with DeserializationUtils methods.


MateuszPol (approved):

modules/swagger-parser/src/main/java/io/swagger/parser/util/DeserializationUtils.java
```diff
@@ -104,12 +103,30 @@ public boolean isYamlAllowRecursiveKeys() {
...
+        public void setMaxYamlCodePoints(Integer maxYamlCodePointsInBytes) {
+            this.maxYamlCodePoints = maxYamlCodePointsInBytes;
+        }
     }
 
     private static Options options = new Options();
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DeserializationUtils.class);
 
+    private static ObjectMapper YAML_MAPPER = Yaml.mapper();
```
frantuma, line 120: Would name it `yamlMapper` or something..

frantuma (commented):
I would probably provide an option to set the `YAMLFactory` instead of the whole mapper, using feature provided by https://github.com/swagger-api/swagger-core/pull/4397

This is because allowing to pass an arbitrary mapper with wrong or absent "Swagger configuration" would result in errors(e.g. when not cloning `Yaml.mapper()` or using `ObjectMapperFactory`) .

That said, I don't see evident issues except the above

CalemRoelofsSB:
@frantuma totally agree, have updated it now to accept a YAMLFactory and create the mapper using it from the ObjectMapperFactory in swagger-core

PR #1937, opened 2023-05-25
Use correct method to read contents of a java.nio.file.Path

makkarpov:
`toFile()` fails when underlying path denotes virtual files (i.e. inside JARs on classpath).
`readAllBytes` works with all filesystems transparently.


ewaostrowska:
Thanks for the PR!  The diagnosis here was correct. However, this issue has since been addressed by [#2320](https://github.com/swagger-api/swagger-parser/pull/2320), which implements the same provider neutral Path handling and includes regression coverage for non-default file systems. 

Since that fix is now merged into master I am closing it as superseded by[ #2320.](https://github.com/swagger-api/swagger-parser/pull/2320)

Issue #1961, opened 2023-08-15
Duplicated schemas produced when shared schema references local schema


AmateurECE:
Consider the following schemas:

TestCase.yaml:
```
components:
  schemas:
    TestCase_TestCase:
      anyOf:
      - $ref: ./TestCase.v1.yaml#/components/schemas/TestCase_v1_TestCase
    TestCase_Foo:
      properties:
        FooTypes:
          items:
            $ref: ./TestCase.yaml#/components/schemas/TestCase_FooType
          type: array
      type: object
    TestCase_FooType:
      enum:
      - All
      - OEM
      type: string
info:
  title: '#TestCase.TestCase'
openapi: 3.0.1
```

TestCase.v1.yaml:
```
components:
  schemas:
    TestCase_v1_TestCase:
      properties:
        Foo:
          $ref: ./TestCase.yaml#/components/schemas/TestCase_Foo
      type: object
info:
  title: '#TestCase.v1.TestCase'
openapi: 3.0.1
```

While parsing TestCase.yaml, `swagger-parser` produces a duplicate schema, `TestCase_Foo_1`. The path of logic that introduces the duplicate schema begins at [ExternalRefProcessor.java:60](https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java#L60), where the comparison fails because `newSchema.properties.get("FooTypes").items.$ref` is `"./TestCase.yaml#/components/schemas/TestCase_FooType"`, while in `existingModel` it's `"#/components/schemas/TestCase_FooType"`. I'm hoping to either bring this issue to light so that the maintainers can implement a fix, or suggest a fix that I can implement and submit a pull request for. I'm open to either!

This behavior is present at the HEAD of master (currently 2.0.17-SNAPSHOT, #08a76302e250), and as far back as 2.0.11. Unfortunately, I can't test with releases pre-2.0.11.

AmateurECE:
Hi there! Has anyone from the team had a chance to review this pull request yet? Thanks very much for your consideration.

PR #1963, opened 2023-08-17
Correctly resolve externalRefs that point to the root path

AmateurECE:
Previously, if OpenAPI v3 document A loads document B, and B refers to schemas in document A, swagger-parser would not resolve references to schemas in A as internal refs, even though they refer to the root OpenAPI v3 document (document A). This would result in the creation of duplicate schemas for those schemas in document A. Now, we resolve these "external" refs as internal refs, preventing the duplication of these schemas by the ExternalRefProcessor.

This fixes issue #1961.


AmateurECE:
Upon further inspection, this only fixes a portion of the failing cases. I've observed two so far--what I'll call "A-B-A," where schemas in B reference schemas in A, and A is the top-level document, but I've discovered a new one--what I'll call "A-B-C-B," where schemas in A reference schemas in B, and schemas in C reference schemas in B. An example schema is given below (not tested yet):

foo.yaml
```
components:
  schemas:
    Group:
      anyOf:
      - $ref: ./bar.yaml#idRef
      - $ref: ./baz.yaml#Group_v1
<...>
```

bar.yaml:
```
components:
  schemas:
    idRef:
      format: uri-reference
<...>
```

baz.yaml:
```
components:
  schemas:
    Group_v1:
      properties:
        id:
          $ref: "./bar.yaml#idRef"
<...>
```

AmateurECE:
I believe this fixes all the issues that I have observed. I have not been able to reproduce the "A-B-C-B" case since I originally saw it, so it's entirely possible that was a fluke. I believe this PR is ready for review/merge.

ewaostrowska:
Hi @AmateurECE! 
Thank you for creating this PR :) the core idea was useful and helped clarify the right approach for https://github.com/swagger-api/swagger-parser/issues/1961. 

We are planning to merge https://github.com/swagger-api/swagger-parser/pull/2383 because it implements that behavior against the current resolver architecture. I have also extended to use URI-based reference identity rather than filesystem Path comparison. That keeps it aligned with the newer normalization work.

PR #1994, opened 2023-10-24
fix parser dereferencing when pathItems are the same

gracekarina:


ttrubinov:
@gracekarina This changes [introduced a new bug](https://github.com/swagger-api/swagger-parser/issues/2037), but didn't fix the bug you are mentioning (you fixed issues when $ref is in responses, but not when $ref in requestBody).
My suggestion is to rollback this PR, I think we can't just delete [that lines](https://github.com/swagger-api/swagger-parser/pull/1994/files#diff-344c61b245c49cbae86308684ac70717a89003cc20535c234f940d5b9f123a12L175-L178)

PR #1997, opened 2023-11-03
feat: introduce cache loaded extensions in OpenAPIDeserializer

### LINKED ISSUE #1909: Cache loaded extensions in OpenApiDeserializer
The ServiceLoader for `JsonSchemaParserExtension`s in `OpenApiDeserializer` is invoked for each detected schema in an OpenAPI document.  The loaded Extensions should be cached statically in `OpenApiDeserializer` (per classloader). 

We are using [openapi-diff](https://github.com/OpenAPITools/openapi-diff) (which in turn uses the `OpenAPIParser`) in our apimangement application to detect api changes. The fact, that

- our application has a large classpath (Spring Boot application)
- the apis to be compared contain many schemas
- the comparison is invoked via rest-service (embedded tomcat) - with a different context classloader, which leads to a second ServiceLoader invocation for each schema

results in poor performance (about 30 seconds) for a comparison because of multiple ServiceLoader invocations.  

This is (admittedly) a special use case. But caching of all detected Extensions would be generally a good idea to avoid unnecessary classpath-scan-operations. 

btw,  there is corresponding `TODO` in `SchemaProcessor`  (https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/SchemaProcessor.java#L49)

harare:
This PR provides a cache mechanism that stores the extensions loaded per classloader.
Due to the caching mechanism the performance for loading the extensions (jsonSchemaParserExtensions) is improved.
This PR closes #1909

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/SchemaProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/SchemaProcessor.java
@@ -47,7 +47,6 @@ public class SchemaProcessor {
             return;
         }
         if (openapi31) {
-            // TODO use as singleton somewhere loaded as static used by both here and deserializer
             List<JsonSchemaParserExtension> jsonschemaExtensions = OpenAPIDeserializer.getJsonSchemaParserExtensions();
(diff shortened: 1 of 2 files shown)


frantuma:
thanks!

Issue #2030, opened 2024-01-04
Swagger Parser doesn't have SafeURLResolve option for OAS 3.0 and OAS 2.0.


MiloszTarka:
The following PRs:

1. https://github.com/swagger-api/swagger-parser/pull/1910
2. https://github.com/swagger-api/swagger-parser/pull/1911

Added a possibility for swagger-parser to provide a Allow & Block list when resolving external refs. Thanks to that, it is possible provide additional security layer when using this tool inside services. This functionality is still missing when resolving 3.0 and 2.0 versions of OAS. 

I've the proposal of the code prepared, so I'll put it on the PR when done.

MiloszTarka:
Addressed in: 
https://github.com/swagger-api/swagger-parser/pull/2032
https://github.com/swagger-api/swagger-parser/pull/2031

frantuma:
implemented in #2031 and #2032

PR #2031, opened 2024-01-04
SWG-9288 utilizing safe url resolver for OAS 2.0

MiloszTarka:
Addresses: https://github.com/swagger-api/swagger-parser/issues/2030

This PR makes swagger-parser to utilizes safeUrlResolver when resolving OpenAPI 2.0 definitions.

The reason behind this change is described deeply under the following PR:
https://github.com/swagger-api/swagger-parser/pull/1910

The $refs are checked in ResolverCache class, in loadRef method.

In order to keep the backward compatibility, I've created a new constructor of ResolverCache so it is possible to explicitly pass the parseOptions through the SwaggerResolver.

### DIFF
--- a/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
+++ b/modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
@@ -13,10 +13,9 @@ import io.swagger.models.properties.Property;
 import io.swagger.models.properties.RefProperty;
 import io.swagger.models.refs.RefFormat;
 import io.swagger.models.refs.RefType;
-import io.swagger.parser.util.DeserializationUtils;
-import io.swagger.parser.util.PathUtils;
-import io.swagger.parser.util.RefUtils;
-import io.swagger.parser.util.SwaggerDeserializer;
+import io.swagger.parser.util.*;
+import io.swagger.v3.parser.urlresolver.PermittedUrlsChecker;
+import io.swagger.v3.parser.urlresolver.exceptions.HostDeniedException;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.File;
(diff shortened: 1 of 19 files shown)


MiloszTarka:
I'll add tests soon & similar PR to v3 parser

modules/swagger-parser/src/main/java/io/swagger/parser/ResolverCache.java
```diff
@@ -66,6 +66,27 @@ public ResolverCache(Swagger swagger, List<AuthorizationValue> auths, String par
         this.swagger = swagger;
```
frantuma, line 66: instead of duplicating logic in constructor I would suggest to call `this(swagger, auths, parentFileLocation, new ParseOptions())`

modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerResolver.java
```diff
@@ -49,6 +50,16 @@ public SwaggerResolver(Swagger swagger, List<AuthorizationValue> auths, String p
         parametersProcessor = new ParameterProcessor(cache, swagger);
     }
 
+    public SwaggerResolver(Swagger swagger, List<AuthorizationValue> auths, String parentFileLocation, Settings settings, ParseOptions parseOptions) {
```
frantuma, line 53: instead of duplicating logic in constructor I would suggest to call in constructor above this(swagger, auths, parentFileLocation, settings, new ParseOptions())

frantuma (changes requested):
Great work, just 2 possible changes in constructors

MiloszTarka:
> Great work, just 2 possible changes in constructors

Great shout, commited changes!

frantuma (approved):

CalemRoelofsSB (approved):

Issue #2037, opened 2024-01-12
Parser incorrectly resolves response $refs, located in same file


ttrubinov:
I have $refs to schemas, that are located in the same file in specification. Latest release of swagger-parser incorrectly resolves responses $refs

Generated OpenAPI should contain `#/components/schemas/` before `MyExampleResponseRef` in path schema $refs, otherwise openapi-generator can't find the schema
However, 2.1.19 release of swagger-parser puts into $refs `#MyExampleResponseRef`, instead of `#/components/schemas/MyExampleResponseRef`

### _Example:_
My  example specification:
**_openapi.yaml:_**
```
openapi: 3.0.3
info:
  title: title
  version: LATEST

paths:
  /get:
    $ref: 'paths/get.yaml#/endpoint'
```
**_get.yaml:_**
```
endpoint:
  get:
    operationId: get
    requestBody:
      content:
        application/json:
          schema:
            $ref: '#/RequestBodyRef'
    responses:
      '200':
        content:
          application/json:
            schema:
              $ref: '#/ResponsesRef'

RequestBodyRef:
  type: string

ResponsesRef:
  type: string
```
OpenAPI output (result of `openApiParser.readLocation(...)`):
<details>
<summary>OpenAPI output</summary>

```
class OpenAPI {
    openapi: 3.0.3
    info: class Info {
        title: title
        description: null
        summary: null
        termsOfService: null
        contact: null
        license: null
        version: LATEST
    }
    externalDocs: null
    servers: [class Server {
        url: /
        description: null
        variables: null
    }]
    security: null
    tags: null
    paths: class Paths {
        {/get=class PathItem {
            summary: null
            description: null
            get: class Operation {
                tags: null
                summary: null
                description: null
                externalDocs: null
                operationId: get
                parameters: null
                requestBody: class RequestBody {
                    description: null
                    content: class Content {
                        {application/json=class MediaType {
                            schema: class Schema {
                                type: null
                                format: null
                                $ref: #/components/schemas/RequestBodyRef
                                description: null
                                title: null
                                multipleOf: null
                                maximum: null
                                exclusiveMaximum: null
                                minimum: null
                                exclusiveMinimum: null
                                maxLength: null
                                minLength: null
                                pattern: null
                                maxItems: null
                                minItems: null
                                uniqueItems: null
                                maxProperties: null
                                minProperties: null
                                required: null
                                not: null
                                properties: null
                                additionalProperties: null
                                nullable: null
[...]

PR #2038, opened 2024-01-15
Fix parser incorrectly resolves response $refs, located in same file 

ttrubinov:
Addresses: [#2037](https://github.com/swagger-api/swagger-parser/issues/2037)

The reason behind this PR described in an issue above and in [this](https://github.com/swagger-api/swagger-parser/pull/1994#issuecomment-1891326691) comment


gracekarina:
Hi, this PR cannot be merged, because that would mean the issues that were trying to get solve will come back, what I can do is take a look at the bug (https://github.com/swagger-api/swagger-parser/issues/2037) generated and fix it.

Issue #2046, opened 2024-01-22
NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "io.swagger.v3.parser.core.models.SwaggerParseResult.getMessages()" is null


garydgregory:
I'll provide a PR to fix:
```
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "io.swagger.v3.parser.core.models.SwaggerParseResult.getMessages()" is null
	at io.swagger.v3.parser.OpenAPIV3Parser.read(OpenAPIV3Parser.java:125)
	at ...
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

PR #2047, opened 2024-01-22
[Issue 2046] NPE in OpenAPIV3Parser.read(String, List<AuthorizationValue>, ParseOptions)

garydgregory:
Issue: https://github.com/swagger-api/swagger-parser/issues/2046
PR: https://github.com/swagger-api/swagger-parser/pull/2047
```
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "io.swagger.v3.parser.core.models.SwaggerParseResult.getMessages()" is null
	at io.swagger.v3.parser.OpenAPIV3Parser.read(OpenAPIV3Parser.java:125)
	at ...
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```
The test is convoluted because the real test would be in the main parser module plus a dependency on this module, which one cannot do due to circularity, but, note that the NPE and fix are "real" and have been tested in production on my end.


gracekarina (approved):

gracekarina:
thank you

garydgregory:
@gracekarina 
Thank you 😊

PR #2054, opened 2024-02-08
[Composite types] When allOf contains only one type don't generate union classes

SergioArrighi:
Hello all,

I don't know if this may be of interest, but I think I saw some issues logged here and there of people asking for this kind of behaviour.
I originally wanted to make it optional, but I did't find any way to have some kind of configuration.

Thanks for the awesome work everybody and best regards.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
@@ -2686,14 +2686,17 @@ public class OpenAPIDeserializer {
 			ComposedSchema composedSchema = new ComposedSchema();
 
 			if (allOfArray != null) {
-
-				for (JsonNode n : allOfArray) {
-					if (n.isObject()) {
-						schema = getSchema((ObjectNode) n, location, result);
-						composedSchema.addAllOfItem(schema);
+				if (allOfArray.size() > 1) {
+					for (JsonNode n : allOfArray) {
+						if (n.isObject()) {
+							schema = getSchema((ObjectNode) n, location, result);
+							composedSchema.addAllOfItem(schema);
+						}
 					}
+					schema = composedSchema;
+				} else if (allOfArray.size() == 1) {
+					schema = getSchema((ObjectNode) allOfArray.get(0), location, result);
 				}
-				schema = composedSchema;
 			}
 			if (anyOfArray != null) {
 


gracekarina:
Hi @SergioArrighi, this change breaks many tests, this could be added as an option if you follow the logic of the resolveCombinators option. thanks

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java
```diff
@@ -2686,14 +2686,17 @@ at the moment path passed as string (basePath) from upper components can be both
 			ComposedSchema composedSchema = new ComposedSchema();
 
 			if (allOfArray != null) {
-
-				for (JsonNode n : allOfArray) {
-					if (n.isObject()) {
-						schema = getSchema((ObjectNode) n, location, result);
-						composedSchema.addAllOfItem(schema);
+				if (allOfArray.size() > 1) {
```
gracekarina, line 2689: if this is going to be a new behavior, you could add it as an option, as the option resolveCombinators

gracekarina (changes requested):
if this is going to be a new behavior, you could add it as an option, as the option resolveCombinators

SergioArrighi:
Hello @gracekarina ,
thanks for the time you took to review this PR. Yes I first wanted to understand if it would be interesting for the project to have this kind of possibility and also to get some guidance on how to have a feature toggle for it. I will look into "resolveCombinators" as soon as I will have some time.
Thanks again for the review and for your work.
Regards

ewaostrowska:
Thanks for the contribution and for explaining the use case @SergioArrighi!

I don't think we should change the parser's default behavior in this way.

This change breaks several existing parser tests and changes the returned model from a ComposedSchema to the inner Schema. That can affect users that depend on the current parser model.

There is also an important difference between:
```

MySchema:
  allOf:
    - $ref: '#/components/schemas/Base'

```
and simply returning Base.

For example, the outer schema may contain additional information:

```
MySchema:
  allOf:
    - $ref: '#/components/schemas/Base'
  description: My special version of Base
  properties:
    extra:
      type: string
```

or it may use another combinator such as `oneOf`/`anyOf`.

If we unwrap the single allOf member while parsing, some of that information can be lost or attached to the wrong schema. The existing regression tests already show problems with descriptions, properties, defaults, references, and composed schemas.

The OpenAPI specification describes allOf as schema composition: the schemas in the array are evaluated together as one composed schema: https://spec.openapis.org/oas/v3.0.4.html#composition-and-inheritance-polymorphism

I think the safer approach is to keep the parser representation unchanged and make this behavior optional.

There is already a similar parser option called resolveCombinators, used together with resolveFully, which controls whether composed schemas are kept or transformed:
https://github.com/swagger-api/swagger-parser/blob/master/README.md#4-resolvecombinators

If the main goal is specifically to avoid an extra generated class, an even better place may be the code generator itself. The generator can recognize a safe single-member allOf and choose not to create an additional class without changing the parser model for everyone.

For these reasons, I'm going to close this PR in its current form.

PR #2058, opened 2024-02-22
Validate the api key 'in' attribute is cookie header or query.

stefan521:
This is a valid API key definition.

```yaml
components:
  securitySchemes:
    ApiKeyAuth:        # arbitrary name for the security scheme
      type: apiKey
      in: header       # can be "header", "query" or "cookie"
      name: X-API-KEY  # name of the header, query parameter or cookie
```

There already is validation that checks the `in` attribute is present. The purpose of this PR is to validate that the attribute value is one of "header", "query", or "cookie".

[Swagger docs](https://swagger.io/docs/specification/authentication/api-keys/) for reference.


stefan521:
Hey @gracekarina, thank you for looking at my PR! 🙂

I updated the existing tests that were failing. The two tests that check the invalid yaml file now got 20 messages instead of 19 because the security schema in the test yaml file looks like this:
` `` securitySchemes:
    api_key:
      type: apiKey
      namex: api_key
      in: not```

frantuma:
Thanks!

PR #2084, opened 2024-04-28
Test and fix #2083 properties in ComposedModel

### LINKED ISSUE #2083: Properties in ComposedModel
[The Azure OpenAPI specification](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2024-03-01/virtualMachine.json#L3321) uses "properties" in definitions that also have `allOf` i.e. are a `ComposedModel`.

I don't know if this is valid or not, but Swagger should support parsing this.

beikov:
Fix #2083


beikov:
Could also check this one please @frantuma?

frantuma:
Thanks!

PR #2093, opened 2024-05-07
fix for issue-2092 - dots lost when dereferencing files with dots

### LINKED ISSUE #2092: dots lost when dereferencing files with dots
Hello

Following fix for https://github.com/swagger-api/swagger-parser/issues/1865 the name of the schema lost the dots

The name of the schema being of importance in the code generation, I think the swagger-parser should keep the dots as per the referenced file name

This mean that at
https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java#L115
I expect to see `#/components/schemas/foo.model`
instead of `#/components/schemas/foomodel`

simonhege:
Fixes https://github.com/swagger-api/swagger-parser/issues/2092

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -137,7 +137,8 @@ public final class ExternalRefProcessor {
                             }
                             schemaFullRef = FilenameUtils.separatorsToUnix(schemaFullRef);
                         }
-                        schema.set$ref(processRefToExternalSchema(schemaFullRef, ref));
+                        // Fix for issue-1865: use internal prefix
+                        schema.set$ref(RefType.SCHEMAS.getInternalPrefix()+ processRefToExternalSchema(schemaFullRef, ref));
                     }
                 } else {
                     processRefToExternalSchema(file + schema.get$ref(), RefFormat.RELATIVE);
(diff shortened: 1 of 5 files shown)


wherka-ama (approved):

wherka-ama:
@char0n , @frantuma  , @kowalczyk-krzysztof, @MiloszTarka, @ponelat : apologies for pulling all of you in here, but we have no idea how to drive this. I'm sorry for the ignorance - it's entirely on me!

Any chance anyone could review this PR please? That would be really appreciated.

Thank you very much in advance! 
Waldek (Amadeus OSPO)

ewaostrowska:
@simonhege Thank you for the contribution! 
This PR cannot be merged yet because even though it fixes the schema components with dots  it breaks unresolved references.

When `processRefToExternalSchema(...) `cannot load a schema, it returns the original external $ref. The new code always adds `#/components/schemas/` so:

`../models/foo.model.yaml`

becomes:

`#/components/schemas/../models/foo.model.yaml`

That points to an internal component that does not exist. This is covered by `testNestedMissingExternalSchemaRetainsOriginalRef.`

The change is also not consistent across component types. Schema references now preserve dots and receive a full internal prefix, but responses, request bodies, parameters, headers, examples, and links can still produce bare references.

For example, an external response may be stored as components.responses.error.response, while its $ref remains:

`error.response
`
instead of:

`#/components/responses/error.response`

These cases are shown in `DottedNonSchemaComponentTest.java`. They are not caused by this PR, but the shared naming change makes the inconsistency more visible.

PR #2096, opened 2024-05-08
OpenAPIDeserializer code cleanup

renatomameli:
No logic changes. Code cleanup only.


taylortails (commented):
looks equivalent to me

renatomameli:
@taylortails Thank you for your feedback. Just to clarify, are you suggesting that the code can be removed because it is equivalent and redundant, or should it remain as it is?

PR #2103, opened 2024-05-30
Update Jackson to v2.17.1

Philzen:
Simple version bump.


Philzen:
BTW – why are there separate properties `${jackson.version}` and `${jackson-databind-version}`? IMHO this is superfluous as all Jackson packages usually are tethered on version bumps.

Issue #1886, opened 2023-02-15
Relative $refs fail on Windows due to potential issue in ExternalRefProcessor#processRefToExternalSchema


kerhard:
Hi,
I have some $ref references in my OpenApi specification that work fine on Linux, but fail on Windows.

The problem seems to be in `ExternalRefProcessor#processRefToExternalSchema` (https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java#L127) within the following code:
```
String parent = (file.contains("/")) ? file.substring(0, file.lastIndexOf('/')) : "";
if (!parent.isEmpty()) {
	if (schemaFullRef.contains("#/")) {
		String[] parts = schemaFullRef.split("#/");
		String schemaFullRefFilePart = parts[0];
		String schemaFullRefInternalRefPart = parts[1];
		schemaFullRef = Paths.get(parent, schemaFullRefFilePart).normalize().toString() + "#/" + schemaFullRefInternalRefPart;
	} else {
		schemaFullRef = Paths.get(parent, schemaFullRef).normalize().toString();
	}
}
```
In the first line the parent folder is determined by cutting the path at the last forward slash.
"openapi/schemas/model.yaml" = > "openapi/schemas"

But the later Paths.get invocation will replace any forward slashes to backslashes on Windows.
Paths.get("openapi/schemas", "ref.yaml").normalize().toString() => "openapi\schemas\ref.yaml"

The replaced path with backslashes seems to then be used for any further embedded references.
So even when I am only using forward slashes in my openapi files, the call to Paths.get will mess up the $ref resolution eventually.

Proposal:
Replace the backslashes with forward slashes after Paths.get is used or make the parent folder determination platform idependent.

adrianhj:
By:

> embedded references

Are you relying on $ref's pointing to items on the classpath where it breaks? Sounds very similar to what I have in #1878 (+ somewhat of a fix in #1879)

i.e. Path is fine when interrogating the Windows filesystem, but once it attempts to retrieve via classpath will always fail.

kerhard:
I am loading from the filesystem, not from the classpath.
[...]

PR #2105, opened 2024-06-17
Normalize relative ref paths to avoid duplicating schemas

### LINKED ISSUE #1518: External ref resolve fails to resolve to same schema and creates duplicate classes
As reported in https://github.com/swagger-api/swagger-parser/issues/1081 and
https://github.com/swagger-api/swagger-parser/issues/1422, something gets confused and creates Foo.java and Foo1.java, even though there is just one foo.json model.

Those issues report that the problem has been fixed, but it has not.

In https://github.com/jimshowalter/swagger-parser-external-ref-bug there is a schema that reproduces the problem.

Run the JavaJAXRSSpecServerCodegen.java code generator on it, and you'll see that it generates a Stunts.java and Stunts1.java, but there is only stunts.json.

Also in that repo is a fix, or at least a fix that works for us. Look for "// BEGIN REFERENCE FORK:".

### LINKED ISSUE #2016: Generating duplicated classes when object is reference more than once
### Generated classes are duplicated in the following situation:
If object A is referenced as a response in API_endpoint1, And object B iis referenced as a response API_endpoint2, where B has a property that references A.
When resolving paths:

- The relatvice path for object **A** resolved from API_endpoint1 looks like this: ./../A.

- The relatvice path for the same object **A** resolved from object B looks like this: ../A.

In this scenario, the presence of both ./ and ../ can lead to the creation of duplicated objects.

ctreatma:
Currently, if an OpenAPI spec contains multiple relative $refs that resolve to the same external resource, but those refs reach that resource through different relative paths, swagger-parser may treat them as distinct references and create duplicate resolved components.

For example, given references such as:

From the spec root: `$ref: ./components/schemas/Thing.yaml`
From a file under components/paths:  `$ref: ../../components/schemas/Thing.yaml`

both references can ultimately identify the same `Thing.yaml` resource, but their paths may contain different relative segments when they reach the resolver. The parser can therefore resolve the same schema more than once, producing components such as `Thing` and `Thing_1` instead of reusing Thing.

This change normalizes URI references before they are used as keys by `ResolverCache`. Equivalent paths containing . or .. segments therefore share the same resolution and external-file cache entries, while the original reference is still used when loading and deserializing the resource.

Because normalization is performed in `ResolverCache`, the behavior applies consistently to reference types resolved through the cache rather than being limited to schemas.

Fixes [#2016](https://github.com/swagger-api/swagger-parser/issues/2016), fixes [#1518](https://github.com/swagger-api/swagger-parser/issues/1518).

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -34,6 +34,7 @@ import io.swagger.v3.oas.models.security.SecurityScheme;
 import io.swagger.v3.parser.ResolverCache;
 import io.swagger.v3.parser.models.RefFormat;
 import io.swagger.v3.parser.models.RefType;
+import io.swagger.v3.parser.util.RefUtils;
 
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -90,6 +91,20 @@ public final class ExternalRefProcessor {
         String renamedRef = cache.getRenamedRef($ref);
         if(renamedRef != null) {
             return renamedRef;
+        }
+
+        RefFormat format = computeRefFormat($ref);
+        if (format.equals(RefFormat.RELATIVE)) {
+            String normalizedRef = "./" + Paths.get($ref).normalize().toString();
+            if (!normalizedRef.equals($ref)) {
+                System.out.println("Normalized " + $ref + " to " + normalizedRef);
+                renamedRef = cache.getRenamedRef($ref);
+                if (renamedRef != null) {
+                    return renamedRef;
+                } else {
+                    $ref = normalizedRef;
+                }
+            }
         }
 
         final Schema schema = cache.loadRef($ref, refFormat, Schema.class);
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -95,7 +95,7 @@ public final class ExternalRefProcessor {
 
         RefFormat format = computeRefFormat($ref);
         if (format.equals(RefFormat.RELATIVE)) {
-            String normalizedRef = "./" + Paths.get($ref).normalize().toString();
+            String normalizedRef = RefUtils.mungedRef(Paths.get($ref).normalize().toString());
             if (!normalizedRef.equals($ref)) {
                 System.out.println("Normalized " + $ref + " to " + normalizedRef);
                 renamedRef = cache.getRenamedRef($ref);
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -34,8 +34,6 @@ import io.swagger.v3.oas.models.security.SecurityScheme;
 import io.swagger.v3.parser.ResolverCache;
 import io.swagger.v3.parser.models.RefFormat;
 import io.swagger.v3.parser.models.RefType;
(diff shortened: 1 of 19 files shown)


modules/swagger-parser-v3/src/test/resources/oas3.fetched/components/parameters/Exclude.yaml
```diff
@@ -0,0 +1,12 @@
+description: |-
```
ctreatma, line 1: In a pinch, I added the impacted spec as a test resource, but I need to trim it down to just the parts necessary for the new test.

nmarriotti:
Just ran into this issue today. Glad to see a fix is on the way (hopefully soon). Thanks!

ctreatma:
I rebased on the latest from `master` to confirm that the issue described in this PR still exists and that the changes in this PR are sufficient to resolve it.

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
```diff
@@ -92,6 +93,20 @@ public String processRefToExternalSchema(String $ref, RefFormat refFormat) {
             return renamedRef;
         }
 
+        RefFormat format = computeRefFormat($ref);
+        if (format.equals(RefFormat.RELATIVE)) {
+            String normalizedRef = "./" + Paths.get($ref).normalize().toString();
```
ctreatma, line 98: To minimize impact on existing tests, this normalizes with a leading `./`.  It might be preferable to leave that out in normalized paths, but that would require changes to a number of tests that currently explicitly check for paths starting with `./`.

ctreatma: It looks like this matches existing behavior so should be OK as-is; [the deserializer adds a leading `./` to relative references](https://github.com/swagger-api/swagger-parser/blob/153070c54acc6b96f902dd9e6155956921453847/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/OpenAPIDeserializer.java#L3935).

Given the existing behavior, an alternative approach to fixing this path resolution issue might be to normalize the relative path in `mungeRef` instead, although I'm not clear if that change would have to happen in only one or in both of `OpenAPIDeserializer.mungeRef` and `RefUtils.mungeRef`.

ctreatma: After further investigation I'm fairly certain this is the only place we can effectively normalize ref paths.  The deserializer can't normalize nested refs and the resolver cache comes into play after we've already created duplicate models.

ewaostrowska:
Hi @ctreatma! Thank you for the PR :) 
I reworked the normalization to keep the fix small and focused.

Instead of using `Paths`, normalization now happens in `ResolverCache` via `URI.normalize()`. The `Paths` approach followed OS filesystem rules, so on Windows it could produce backslashes like `..\schemas\Foo.yaml`. OpenAPI `$ref` values are URI references and must use `/`, so that was a correctness issue on Windows regardless of the duplicate-schema problem you were fixing.

The original `$ref` string is preserved for loading and error reporting — only the cache key is normalized. Because canonicalization sits in `ResolverCache`, it applies to every `loadRef` caller: `PathsProcessor`, `OperationProcessor`, `ParameterProcessor`, and `SecuritySchemeProcessor`, not just schemas. The cache is the single point where ref identity is decided, so normalizing there guarantees consistent behavior regardless of which processor resolves the ref first.

I expanded the regression tests to cover [#2016](https://github.com/swagger-api/swagger-parser/issues/2016), [#1518](https://github.com/swagger-api/swagger-parser/issues/1518), [#1886](https://github.com/swagger-api/swagger-parser/issues/1886), URI fragments, URL and file refs, invalid URI fallback, and a non-schema `ApiResponse` case.

daniel-kmiecik (approved):

ewaostrowska:
Updated the public caches to keep their original ref/file keys, with separate normalized caches added internally for deduplication. Consumers using getResolutionCache(), getExternalFileCache(), or getRenameCache() see no change; equivalent refs now share the same internal cache identity. Tests cover both cases.

Issue #2113, opened 2024-07-30
Absent type doesn't imply type object


wsalembi:
https://github.com/swagger-api/swagger-parser/commit/9a5cd1990b06374c75999288824dca83a2cef4e3 introduced the behavior that properties without type defined are implied of type object. I don't think this statement is true. The absence of type means it is any type (object, null, string, etc).

Example

```
paths:
  /foo:
    post:
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Foo'
        required: true
      responses:
        200:
          description: ok
components:
  schemas:
    Foo:
      type: object
      allOf:
        - $ref: "#/components/schemas/Goo"
    Goo:
      type: object
      properties:
        goo:
          title: "Goo"
````

When resolving the schema, the type of property `goo` cannot be set to type object.

I removed the implication and it didn't break any other tests.

```
diff --git a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java	(revision 0ab2e8a7774b147e7b8c2c75b929089677d1b246)
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java	(date 1722333547892)
@@ -469,13 +469,13 @@
                 Schema property = updated.get(key);
 
                 if (property.getProperties() != model.getProperties()) {
-                    if (!hasSchemaType(property)) {
+                    /*if (!hasSchemaType(property)) {
                         if (SpecVersion.V30.equals(property.getSpecVersion())) {
                             property.setType("object");
                         } else {
                             property.addType("object");
                         }
-                    }
+                    }*/
                     model.addProperties(key, property);
                 } else {
                     LOGGER.debug("not adding recursive properties, using generic object");
```

jeremyfiel:
You have one incorrect assumption here. 

The `type` keyword does not support `null` in OAS 3.0.x

You must use `nullable` as an alternative  

> null is not supported as a type (see [nullable](https://spec.openapis.org/oas/v3.0.3.html#schemaNullable) for an alternative solution)

Otherwise, I agree you cannot force `type: object` where any other primitive is a valid schema.

wsalembi:
[...]

PR #2114, opened 2024-07-30
Absent type doesn't imply type object #2113

wsalembi:
Fixes issue #2113 
+ some fixes so test run also on Windows


wsalembi:
@gracekarina @frantuma Can you please review the fix and merge into the master branch? Thank you!

PR #2123, opened 2024-10-04
fix(#2104): properly process inline refs in composed schemas

### LINKED ISSUE #2104: ExternalRefProcess.processComposedSchema not processing inline schema definitions for oneOf's and anyOf's
Hi, 
I'm using the following version of the swagger-parser-v3 to resolve my api.
<dependency>
            <groupId>io.swagger.parser.v3</groupId>
            <artifactId>swagger-parser-v3</artifactId>
            <version>2.1.22</version>
        </dependency>

When referencing schemas in oneOf's inline schema definitions properties, the resolved yaml file doesn't contain the referenced schema definitions. 

image

For the complete example, see the content of the following zip. 
[openapi.zip](https://github.com/user-attachments/files/15864606/openapi.zip)

This is due to a duplicated conditional line in the class ExternalRefProcess and method processComposedSchema. 
image

The same duplicated line is present for the anyOfs and is probably incorrect as well.

fpaul-1A:
Fix #2104


frantuma:
Thanks!

PR #2127, opened 2024-10-16
Enhancement/parser resolve responses flag

micryc:
Due to unexpected results during parsing OAS with options - resolve and flatten set to true, we decided to create resolveResponses flag, from now responses will not be fully resolved by default with option resolve - true.
If anyone wants to keep previous behaviour needs to use option `setResolveResponses(true)`

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java
@@ -70,7 +70,7 @@ public class OperationProcessor {
             for (String responseCode : responses.keySet()) {
                 ApiResponse response = responses.get(responseCode);
                 if(response != null) {
-                    //This part allows parser to put response inline without the resolveFully option set to true
+                    //This part allows parser to put response inline when resolveResponses = true
                     if (response.get$ref() != null && cache != null && cache.getParseOptions() != null && cache.getParseOptions().isResolveResponses()) {
 
                         responseProcessor.processResponse(response);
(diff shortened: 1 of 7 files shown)


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java
```diff
@@ -70,8 +70,8 @@ public void processOperation(Operation operation) {
             for (String responseCode : responses.keySet()) {
                 ApiResponse response = responses.get(responseCode);
                 if(response != null) {
-                    //This part allows parser to put response inline without the resolveFully option set to true
-                    if (response.get$ref() != null) {
+                    //This part allows parser to put response inline within resolveResponses or ResolveFully
```
frantuma, line 73: would replace with `// This part allows parser to put response inline when resolveResponses = true`

see below

modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/OperationProcessor.java
```diff
@@ -70,8 +70,8 @@ public void processOperation(Operation operation) {
...
+                    if (response.get$ref() != null && cache != null && cache.getParseOptions() != null && (cache.getParseOptions().isResolveResponses() || cache.getParseOptions().isResolveFully())) {
```
frantuma, line 74: this shouldn't be triggered by `resolveFully` but only by the flag. It's true that we currently don't process `components/responses` in `ResolveFully` as we do for requestBodies, but this should be addressed in separate ticket and logic added to `ResolveFully` and not here

frantuma (changes requested):

frantuma (approved):

Issue #2102, opened 2024-05-16
Parameters with the same name but different locations are incorrectly returned when using JSON references ($ref) with openapi 3.0.x.


AdrianVasiliu:
Multiple parameters with the same name but different locations (for instance, `in: query` vs `in: header`) are wrongly handled by the parser if:
* the parameters are described using JSON references (`$ref`).
* and the OpenAPI file declares its version as v3.0.x (`"openapi": "3.0.0"` or `"openapi": "3.0.3"`).

The issue does NOT occur when either not using `$ref`, or declaring `"openapi": "3.1.0"`.

This looks like a bug because:

a) It violates the OpenAPI spec (both v3.0 and v3.1):

> A unique parameter is defined by a combination of a [name](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#parameterName) and [location](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#parameterIn).

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#parameter-object

This should be respected regardless of whether the parameter is defined with, or without `$ref`.

b) I don't see a mention of a change in this area between v3.0 and v3.1 (for instance [here](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0) or [here](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---OpenAPI-3.1)) which would justify the difference of behaviour.

---

How to reproduce:

```java
package parsertest;

import io.swagger.v3.oas.models.OpenAPI;
import io.swagger.v3.oas.models.parameters.Parameter;
import io.swagger.v3.parser.OpenAPIV3Parser;
import io.swagger.v3.parser.core.models.ParseOptions;

import java.io.IOException;
import java.util.List;

public class ParserTest {
    public static void main(String[] args) throws IOException {
        test("src/test/resources/parsertest/openapi-ok.json");
        test("src/test/resources/jparsertest/openapi-ko.json");
    }
    
    private static void test(String openapiFile) throws IOException {
        System.out.println("=== OpenAPI file:" + openapiFile);
        final ParseOptions options = new ParseOptions();
        options.setResolve(true);
        options.setResolveCombinators(true);
        options.setResolveFully(true);
        
        final OpenAPI openAPI = new OpenAPIV3Parser().read(openapiFile, null, options);
        
[...]

PR #2250, opened 2025-11-21
fix: same name and different location in param + small refactor

daniel-kmiecik:
Fix for #2102

### DIFF
--- a/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/ParserDeduplicationParameterTest.java
+++ b/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/ParserDeduplicationParameterTest.java
@@ -12,7 +12,7 @@ import java.util.stream.Collectors;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertTrue;
 
-public class ParserDeduplicationTest {
+public class ParserDeduplicationParameterTest {
 
     @Test
     public void testIssue2102TwoRefsSameNameDifferentIn31() {


modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ParameterProcessor.java
```diff
@@ -33,137 +32,140 @@ public ParameterProcessor(ResolverCache cache, OpenAPI openAPI) {
...
+                if (mediaType.getSchema() != null) {
+                    Optional<Schema> schema = Optional.ofNullable(mediaType.getSchema());
+                    schema.ifPresent(schemaProcessor::processSchema);
                 }
             }
         }
     }
 
-   public List<Parameter> processParameters(List<Parameter> parameters) {
+    public List<Parameter> processParameters(List<Parameter> parameters) {
```
ewaostrowska: Please consider breaking this method into even smaller ones

modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/OpenAPIV3ParserTest.java
```diff
@@ -3436,4 +3437,17 @@ public void testVersion(){
         SwaggerParseResult parseResult = openApiParser.readLocation("version-missing.yaml", null, options);
         assertEquals(parseResult.getMessages().get(0), "attribute info.version is missing");
     }
+
```
ewaostrowska: Please consider adding few similar tests: 
- for 3.1 
- with different resolution options (such as ones in the original issue for example)
- with mixed inlined and ref params of the same name and different location 
- testing other options of combination (name, in)

ewaostrowska (changes requested):

PR #2294, opened 2026-03-23
Fix thread-safety bug in OpenAPIDereferencer31

### LINKED ISSUE #2293: Thread-safety bug: OpenAPIDereferencer31 singleton has mutable instance fields overwritten on each call
## Description

`OpenAPIDereferencer31` is instantiated as a singleton via `DereferencersFactory`, but its `dereference()` method stores per-call state in mutable instance fields:

```java
// OpenAPIDereferencer31.java lines 20-21
private OpenAPI openAPI;
private SwaggerParseResult result;

// lines 41-42 in dereference()
openAPI = context.openApi;
result = context.swaggerParseResult;
```

When multiple threads parse different OpenAPI 3.1 specs concurrently, these fields get overwritten by whichever thread runs last, causing one parse to return another parse's result.

## Steps to Reproduce

Parse two different OpenAPI 3.1 specs concurrently:

```java
import io.swagger.parser.OpenAPIParser;
import io.swagger.v3.parser.core.models.ParseOptions;
import java.util.concurrent.*;
import java.util.stream.*;

public class ConcurrencyBug {
    static String makeSpec(String title) {
        return """
            {
              "openapi": "3.1.0",
              "info": { "title": "%s", "version": "1.0.0" },
              "paths": {},
              "components": {
                "schemas": {
                  "Item": {
                    "type": "object",
                    "properties": { "id": { "type": "string" } }
                  },
                  "ItemList": {
                    "type": "object",
                    "properties": {
                      "items": {
                        "type": "array",
                        "items": { "$ref": "#/components/schemas/Item" }
                      }
                    }
                  }
                }
              }
            }
            """.formatted(title);
    }

    static String parseWithResolve(String spec) {
        options.setResolve(true);
        options.setResolveFully(false);
        return new OpenAPIParser().readContents(spec, null, options)
            .getOpenAPI().getInfo().getTitle();
    }

    public static void main(String[] args) throws Exception {
        String specA = makeSpec("SpecA");
        String specB = makeSpec("SpecB");

        ExecutorService pool = Executors.newFixedThreadPool(8);
        for (int round = 0; round < 10; round++) {
            List<Future<String>> futures = IntStream.range(0, 200)
                .mapToObj(i -> pool.submit(() -> parseWithResolve(i % 2 == 0 ? specA : specB)))
                .collect(Collectors.toList());

            for (int i = 0; i < futures.size(); i++) {
                String expected = (i % 2 == 0) ? "SpecA" : "SpecB";
                String actual = futures.get(i).get();
                if (\!expected.equals(actual)) {
                    System.out.println("CORRUPTED: expected=" + expected + " actual=" + actual);
                }
            }
        }
        pool.shutdown();
        System.out.println("Done");
    }
}
```

## Expected Behavior

Each parse returns the correct `OpenAPI` object matching its input spec.

## Actual Behavior

Concurrent parses of OpenAPI 3.1 specs return each other's results. For example, parsing SpecA returns SpecB's `OpenAPI` object.

This only affects OpenAPI 3.1 specs because the 3.1 code path goes through the singleton `OpenAPIDereferencer31`, while 3.0 specs use a per-call `OpenAPIResolver` instance.

## Root Cause

`DereferencersFactory` creates a single `OpenAPIDereferencer31` instance (line 20 of `DereferencersFactory.java`), and `OpenAPIDereferencer31.dereference()` writes to instance fields `openAPI` and `result` on each call. Concurrent calls overwrite each other's state.

## Suggested Fix

Make `openAPI` and `result` local variables in `dereference()` instead of instance fields. They are only used within that method and don't need to be stored on the instance:

```java
public void dereference(DereferencerContext context, Iterator<OpenAPIDereferencer> chain) {
    if (\!canDereference(context)) {
        if (chain.hasNext() && chain.next().canDereference(context)) {
            chain.next().dereference(context, chain);
            return;
        }
    }

    OpenAPI openAPI = context.openApi;              // local variable
    SwaggerParseResult result = context.swaggerParseResult;  // local variable
    // ... rest of method unchanged
}
```

## Version

swagger-parser 2.1.39

dontgitit:
## Summary

Fixes #2293

`OpenAPIDereferencer31` is instantiated as a singleton via `DereferencersFactory`, but its `dereference()` method stored per-call state in mutable instance fields (`openAPI`, `result`, `messages`). When multiple threads parse different OpenAPI 3.1 specs concurrently with `resolve=true`, these fields get overwritten by whichever thread runs last, causing one parse to return another parse's result.

**Fix:** Make `openAPI` and `result` local variables in `dereference()` instead of instance fields, since they are only used within that method. Also remove the unused `messages` field which had the same issue.

This only affects OpenAPI 3.1 specs — the 3.0 code path uses a per-call `OpenAPIResolver` instance and is not affected.

## Changes

- `OpenAPIDereferencer31.java`: Converted `openAPI` and `result` from instance fields to local variables in `dereference()`. Removed unused `messages` field.
- `OpenAPIDereferencer31ThreadSafetyTest.java`: New test that parses two different OpenAPI 3.1 specs concurrently (200 iterations across 8 threads) and asserts each result matches its input.

## Test plan

- [x] New test `OpenAPIDereferencer31ThreadSafetyTest` fails without the fix, passes with it
- [x] Verified by temporarily reverting the fix and running the test — fails with `ExecutionException` due to assertion mismatch


abpurdy-google (approved):

djankows (approved):

djankows:
Hi @dontgitit, appreciate the contribution - will be part of next release.

PR #2343, opened 2026-06-09
Validate empty Swagger property names

### LINKED ISSUE #2339: [Bug]: Anonymous objects fail to report errors during parsing
## Description

The below Swagger doc passes through the SwaggerParser without generating any error messages. The same doc opened in various code editors [including editor.swagger.io] highlights the issue with the lack of object name.  In addition, duplicate object names do not seem to be parsed so setting both objects to a name of "" overwrites the first definition with the second silently - "" being a valid string in Java, even if not valid according to the Swagger spec.

## Affected Version

1.0.71
Also using SwaggerConvertor and 2.1.22, the converted doc doesn't flag an issue.

## Steps to Reproduce

```
{
    "swagger": "2.0",
    "info": {
        "title": "info_title",
        "version": "0/0",
        "description": "..",
        "contact": {
            "name": "contactname",
            "url": "contacturl",
            "email": "contact@email.com"
        }
    },
    "produces": [
        "application/json"
    ],
    "consumes": [
        "application/json"
    ],
    "basePath": "/basepath",
    "paths": {
        "/path": {
            "get": {
                "summary": "summary",
                "responses": {
                    "200": {
                        "description": "..",
                        "schema": {
                            "type": "object",
                            "properties": {
                                "AA": {
                                    "type": "string"
                                },
                                "": {
                                  "type": "number"
                                }
                            }
                        }
                    }
                },
                "operationId": "getopid",
                "description": ".."
            }
        }
    }
}
```

## Expected Behavior

Error messages similar to those from editor.swagger.io:
```
properties members must be schemasapilint(10037)
properties: object
The value of "properties" MUST be an object. Each value of this object MUST be a valid JSON Schema.
```

## Actual Behavior

Document parsed successfully; attempts to reference the "anonymous" fields cause further processing issues

## Environment

- Java version: various
- Build tool: maven
- OS: Fedora 43

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-parser/issues) and this is not a duplicate.

king-407:
Fixes #2339.

## What changed

This adds validation for Swagger 2.0 schemas that contain an empty property name under `properties`.

Previously, a schema like this was parsed without reporting a validation message:

```json
"properties": {
  "AA": { "type": "string" },
  "": { "type": "number" }
}
```

The converter now adds a validation message for the empty property name while continuing to parse the document.

## Testing

Ran the focused regression test and the full `V2ConverterTest` class:

```bash
mvn -pl modules/swagger-parser-v2-converter -am -Dtest=V2ConverterTest#testEmptyPropertyNameHasMessage -Dsurefire.failIfNoSpecifiedTests=false test
mvn -pl modules/swagger-parser-v2-converter -am -Dtest=V2ConverterTest -Dsurefire.failIfNoSpecifiedTests=false test
```

### DIFF
--- a/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
+++ b/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java
@@ -153,6 +153,7 @@ public class SwaggerConverter implements SwaggerParserExtension {
         SwaggerInventory inventory = new SwaggerInventory().process(parse.getSwagger());
 
         Swagger swagger = parse.getSwagger();
+        addEmptyPropertyNameMessages(swagger, output);
 
         if (swagger.getVendorExtensions() != null) {
             openAPI.setExtensions(convert(swagger.getVendorExtensions()));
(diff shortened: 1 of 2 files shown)


ewaostrowska:
Thanks for the fix and the regression test!
We looked at the relevant specs and don't think an empty key in properties is a Swagger 2.0 validation error.

The [OpenAPI 2.0 Schema Object](https://spec.openapis.org/oas/v2.0.html#schema-object) applies JSON Schema semantics to properties. [JSON Schema Draft 4](https://json-schema.org/draft-04/json-schema-validation#section-5.4.4.1) requires properties to be an object where each value is a valid schema, but says nothing about the names being non-empty. The official [OpenAPI 2.0 validation schema](https://spec.openapis.org/oas/2.0/schema/2017-08-27.html) has no such restriction either. 

Duplicate names are a separate problem. RFC 8259 recommends unique names and warns that behavior is unpredictable otherwise, but by the time SwaggerConverter receives a Map<String, Property>, duplicates have already been collapsed. A converter traversal can't reliably detect that data loss.

Emitting a warning here would flag spec-valid documents as broken and could cause consumers that treat parser messages as errors to reject those documents. 

We're closing this without merging. Thanks again for digging into this!

PR #2349, opened 2026-06-13
Resolve same-named schemas from different files as distinct components (fixes #2333)

### LINKED ISSUE #2333: [Bug]: Parser confuses referenced models with same name
## Description

When parsing an openapi spec (using resolve: true) that has models A referencing C from file1 and B referencing C from file2, it will use C from file1 for both A and B.

I have created a reproduction in openapi-generator here: https://github.com/OpenAPITools/openapi-generator/commit/43f0c76683232e8c670cace664e42624ba3d3294, only noticing afterwards, that the bug lies with swagger-parser.

The downstream project uses the following code for parsing:

https://github.com/OpenAPITools/openapi-generator/blob/81115a094b2c80f9d23bc5a28d717d65ed043177/modules/openapi-generator/src/main/java/org/openapitools/codegen/config/CodegenConfigurator.java#L683-L687

```java
        final List<AuthorizationValue> authorizationValues = AuthParser.parse(this.auth);
        options.setResolve(true);
        options.setResolveResponses(true);
        SwaggerParseResult result = new OpenAPIParser().readLocation(inputSpec, authorizationValues, options);
```

## Affected Version

2.1.41

## Steps to Reproduce

Try to parse the following spec using resolve: true.

### Main spec file

```yaml
openapi: 3.0.3
info:
  title: Pet Store API
  version: 1.0.0
paths: {}
components:
  schemas:
    SomeItem:
      $ref: "inventory.yaml#/components/schemas/Pet"
    Pet:
      $ref: "pets.yaml#/components/schemas/Pet"
```

### inventory.yaml

```yaml
openapi: 3.0.3
info:
  title: Inventory API
  version: 1.0.0
paths: {}
components:
  schemas:
    Pet:
      type: object
      properties:
        name:
          type: string
```

### pets.yaml

```yaml

openapi: 3.0.3
info:
  title: Pets API
  version: 1.0.0
paths: {}
components:
  schemas:
    Pet:
      type: object
      properties:
        id:
          type: integer
```

## Expected Behavior

`SomeItem` should include a `name` property and `Pet` should include an `id` property.

## Actual Behavior

Both `SomeItem` and `Pet` use the same model `Pet` from `inventory.yaml` and contain the `name` attribute.

## Environment

- Java version: OpenJDK 11
- Build tool: mvn 3.9.15
- OS: Arch Linux

## Checklist

- [x] I have searched the [existing issues](https://github.com/swagger-api/swagger-parser/issues) and this is not a duplicate.

seonwooj0810:
### Problem
When parsing a spec with `resolve: true` where two schemas in **different files share the same name**, both references collapse into the first model. In the issue's reproducer, `SomeItem` (→ `inventory.yaml#/components/schemas/Pet`, has `name`) and `Pet` (→ `pets.yaml#/components/schemas/Pet`, has `id`) both end up pointing at inventory's `Pet`.

### Root cause
`ExternalRefProcessor.finalNameRec` decides which name a freshly resolved external schema gets. When the candidate name is already taken by an entry that still carries a `$ref` (an unresolved external placeholder, e.g. the sibling `Pet` component pointing at `pets.yaml`), the code unconditionally treated that slot as free and reused the name — then `addSchemas(newRef, schema)` overwrote the placeholder. The distinct second schema was lost.

### Fix
A `$ref` placeholder slot is only safe to reuse when it targets **the same reference** currently being resolved. The incoming `$ref` is now threaded into `finalNameRec`; if the existing placeholder points to a different target, resolution falls through to the next suffixed name (`Pet_1`) instead of clobbering it. This is the conservative direction — a mismatch only ever produces an extra suffixed name, never data loss.

### Result on the reproducer
```
SomeItem -> { name }     (inventory.yaml Pet)
Pet      -> { id }       (pets.yaml Pet)
Pet_1    -> { name }
```
`SomeItem` correctly carries `name` and `Pet` correctly carries `id`.

### Test evidence
- Added `Issue2333Test` with the issue's three-file reproducer asserting the two `Pet` schemas resolve distinctly. Passes.
- Full `swagger-parser-v3` module suite: **609 tests, 0 failures, 0 errors** (run with `-Dmaven.javadoc.skip=true`; the offline javadoc fetch and the unrelated `swagger-parser-safe-url-resolver` `PermittedUrlsCheckerTest` failures are pre-existing JDK 25 environment issues, not related to this change).

### Verification done
1. No in-flight PR (`gh pr list` search for the issue/keywords — none open); 2. no self-claim in the thread; 3. code-focused (`.java`); 4. reproduced the bug on `master`, confirmed the fix flips the behavior; 5. confirmed root symbol `finalNameRec`/`processRefToExternalSchema` still present on `master`; 6. n/a (no parent epic).

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/ResolverCache.java
@@ -38,6 +38,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -420,6 +421,10 @@ public class ResolverCache {
     public void putRenamedRef(String originalRef, String newRef) {
         renameCache.put(originalRef, newRef);
         canonicalRenameCache.put(canonicalize(originalRef), newRef);
+    }
+
+    public boolean refsAreEquivalent(String first, String second) {
+        return Objects.equals(canonicalize(first), canonicalize(second));
     }
 
     private static String canonicalize(String ref) {
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ExternalRefProcessor.java
@@ -57,11 +57,24 @@ public final class ExternalRefProcessor {
 
     private String finalNameRec(Map<String, Schema> schemas, String possiblyConflictingDefinitionName, Schema newSchema,
         int iteration) {
+        return finalNameRec(schemas, possiblyConflictingDefinitionName, newSchema, iteration, null);
+    }
+
+    private String finalNameRec(Map<String, Schema> schemas, String possiblyConflictingDefinitionName, Schema newSchema,
+        int iteration, String incomingRef) {
         String tryName =
             iteration == 0 ? possiblyConflictingDefinitionName : possiblyConflictingDefinitionName + "_" + iteration;
         Schema existingModel = schemas.get(tryName);
         if (existingModel != null) {
             if (existingModel.get$ref() != null) {
+                // The name is currently held by a not-yet-resolved external $ref placeholder. It is only safe to
+                // reuse the name when that placeholder targets the same reference we are resolving; otherwise the
+                // placeholder is a distinct schema that would be silently overwritten (see issue #2333).
+                if (incomingRef != null && existingModel.get$ref() != null
+                        && !existingModel.get$ref().equals(incomingRef)) {
+                    LOGGER.debug("A different external $ref already claims the name " + tryName);
+                    return finalNameRec(schemas, possiblyConflictingDefinitionName, newSchema, ++iteration, incomingRef);
+                }
                 // use the new model
                 existingModel = null;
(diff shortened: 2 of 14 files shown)


seonwooj0810:
Gentle ping — open since mid-June.

Recap of #2333: with `resolve: true`, two schemas with the same name in different files collapse into the first model, because `ExternalRefProcessor.finalNameRec` treats an existing entry that still carries a `$ref` (an unresolved external placeholder) as a match and reuses its name. The PR keeps them distinct.

Is there anything you'd like adjusted before this can be considered — naming strategy for the second component in particular? Still merges cleanly and both checks are green.

ewaostrowska:
Hi @seonwooj0810! Thank you for the PR :)

Since this PR was opened, [#2105](https://github.com/swagger-api/swagger-parser/pull/2105) introduced canonical $ref identity in `ResolverCache`. I've rebased the local changes on that, so the fix uses the same normalization instead of raw string equality. That keeps distinct refs separate and deduplicates equivalent paths containing `.` or `...` 

The regression tests cover both behaviors to stay compatible with [#2105.](https://github.com/swagger-api/swagger-parser/pull/2105)

daniel-kmiecik (approved):

djankows (approved):

ewaostrowska:
One note here: This PR fixes the collision for `components/schemas` only. The same problem exists in `components/responses` and `components/parameters` too, most likely others as well, but fixing all component types needs a shared naming mechanism and more tests. Better to keep this PR focused on the reported case and open a separate issue for the rest.

PR #2363, opened 2026-07-02
Test operation security alternatives

kriptoburak:
## Summary

- Add a parser regression for operation security with an anonymous alternative and an API key requirement.
- Assert both operation-level security requirement entries are preserved.

## Validation

- Static whitespace check passed before commit.
- Not run: this environment does not have a Java runtime or Maven.


kriptoburak:
Rebased onto the current `master` branch at `4467854473a2a3a8b70413aeecaa279100491d5f`. `git range-diff` confirms the regression patch is unchanged.

Local validation with JDK 21:

- `OpenAPIDeserializerTest`: 89 tests passed
- Reactor `verify` passed for `swagger-parser-v3` and dependencies
- `git diff --check` passed

This remains a test-only change.

PR #2366, opened 2026-07-08
fix: normalize raw local path base URI for OpenAPI 3.1 ref resolution (#2365)

### LINKED ISSUE #2365: [Bug]: OpenAPI 3.1 same-file $ref resolution reports illegal path when raw local path contains spaces
## Description

`OpenAPIV3Parser.readLocation(...)` parses an OpenAPI 3.1 document from a raw local filesystem path containing spaces, but same-file `$ref` resolution records an `Illegal character in path` error.

The spec is a single file. It does not use sibling or external spec files.

The failing case is a response `$ref` within the same OpenAPI document:

```yaml
paths:
  /examples:
    get:
      responses:
        '200':
          $ref: '#/components/responses/ExampleResponse'
components:
  responses:
    ExampleResponse:
      description: OK
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/Example'
```

When the file is parsed from a raw path like:

```text
/tmp/openapi path with spaces.../openapi.yaml
```

the parser reports `Illegal character in path`.

## Affected Version

Current `master` / `2.1.46-SNAPSHOT`.

Earliest version the bug appears in (if known):  
I have also reproduced similar raw-path-with-spaces behavior against released 2.1.x versions, so this does not appear to be a recent regression.

## Steps to Reproduce

1. Create an OpenAPI 3.1 file in a directory whose path contains spaces.
2. Include a same-file response `$ref`, such as `#/components/responses/ExampleResponse`.
3. Parse it with `OpenAPIV3Parser.readLocation(...)` using the raw filesystem path.

Minimal repro branch:

https://github.com/salockhart/swagger-parser/tree/feat/openapi31-space-path-ref-repro

Relevant commit:

```text
d47f782d test: reproduce OpenAPI 3.1 path with spaces ref issue
```

Focused test command:

```bash
mvn -pl modules/swagger-parser-v3 -am \
  -Dtest=OpenAPIV31ParserUriTest#resolveSameFileReferenceFromRawPathWithSpaces \
  -DfailIfNoTests=false \
  -Dsurefire.failIfNoSpecifiedTests=false \
  test
```

## Expected Behavior

A raw local path accepted by `OpenAPIV3Parser.readLocation(...)` should not produce an `Illegal character in path` error while resolving same-file OpenAPI 3.1 references.

Alternatively, if raw local paths are not intended to be supported when they contain spaces, the parser should fail clearly or document that callers must pass a URI-safe file URL.

## Actual Behavior

The parser returns an OpenAPI model, but the parse result contains an error message like:

```text
Illegal character in path at index ...: /tmp/openapi path with spaces.../openapi.yaml
```

The error is emitted while resolving same-file OpenAPI 3.1 refs.

## Logs / Stack Traces

The focused repro test currently fails with:

```text
java.lang.AssertionError: expected [null] but found [Illegal character in path at index ...: /tmp/openapi path with spaces.../openapi.yaml]
```

The stack trace points through OpenAPI 3.1 reference resolution, including:

```text
io.swagger.v3.parser.reference.Visitor.readURI
io.swagger.v3.parser.reference.ReferenceVisitor.resolveRef
io.swagger.v3.parser.reference.ReferenceVisitor.visitResponse
io.swagger.v3.parser.reference.OpenAPIDereferencer31.dereference
io.swagger.v3.parser.OpenAPIV3Parser.readLocation
```

## Environment

- Java version: reproduced locally with Temurin 25.0.3
- Build tool: Maven
- OS: macOS

## Additional Context

Adjacent but not exact existing issues:

- https://github.com/swagger-api/swagger-parser/issues/1572
- https://github.com/swagger-api/swagger-parser/issues/2159
- https://github.com/swagger-api/swagger-parser/issues/2136

## Checklist

- [x] I have searched the existing issues and this is not a duplicate.

seonwooj0810:
Fixes #2365

## Root cause

`OpenAPIV3Parser.readLocation(...)` accepts a raw local filesystem path, and that path becomes the base URI for OpenAPI 3.1 reference resolution. Resolution then builds `java.net.URI` instances from the base in several places (`ReferenceUtils.resolve/toBaseURI` and `Visitor#readURI`). When the raw path contains a character that is illegal in a URI — most commonly a space — those `new URI(...)` calls throw `URISyntaxException: Illegal character in path`. The exception is caught in `ReferenceVisitor` and recorded as a parse message, so a self-contained single-file spec resolving same-file refs (e.g. `#/components/responses/...`) returns a result polluted with an `Illegal character in path` error.

## Fix

Normalize the location once, at the entry to `resolve(...)`: if the location is not already a valid URI, convert it to a proper `file:` URI via `Paths.get(location).toUri()` so the base is URI-safe everywhere downstream. Locations that already parse as a URI (including scheme-less relative paths without illegal characters) are left untouched, so existing behavior is unchanged.

## Test evidence

Added a regression test `OpenAPIV31RawPathWithSpacesTest#resolveSameFileReferenceFromRawPathWithSpaces` that writes a single-file 3.1 spec with same-file response/schema `$ref`s into a temp directory whose path contains spaces and asserts the parse result carries no `Illegal character in path` message. It fails before the change (`Illegal character in path at index 12: ...`) and passes after.

Verification done: full `swagger-parser-v3` module test suite green (584 + 25 tests, 0 failures) with the fix; the new regression test fails before / passes after. (The `safe-url-resolver` module has pre-existing JDK 25 mock failures unrelated to this change.)

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/OpenAPIV3Parser.java
@@ -21,6 +21,7 @@ import io.swagger.v3.parser.util.RemoteUrl;
 import io.swagger.v3.parser.util.ResolverFully;
 import java.io.InputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -209,6 +210,7 @@ public class OpenAPIV3Parser implements SwaggerParserExtension {
             String location) {
         if (location != null) {
             location = location.replace('\\', '/');
+            location = toUriSafeLocation(location);
         }
         try {
             if (options != null) {
@@ -263,6 +265,29 @@ public class OpenAPIV3Parser implements SwaggerParserExtension {
             // result.setMessages(Collections.singletonList(e.getMessage()));
         }
         return result;
+    }
+
+    /**
+     * Normalizes a location that is used as the base URI for reference resolution.
+     * <p>
+     * {@code readLocation} accepts raw local filesystem paths, but reference resolution builds
+     * {@code java.net.URI} instances from that base (e.g. in {@code ReferenceUtils} and {@code Visitor#readURI}).
+     * A raw path containing characters that are illegal in a URI (most commonly a space) makes those
+     * constructions fail with "Illegal character in path", which surfaces as a parse error even for a
+     * self-contained single-file spec. When the location is not already a valid URI, convert it to a proper
+     * {@code file:} URI so the base is URI-safe; valid URIs (including scheme-less relative paths) are left untouched.
+     */
+    private static String toUriSafeLocation(String location) {
+        try {
+            new URI(location);
+            return location;
+        } catch (URISyntaxException e) {
+            try {
+                return Paths.get(location).toUri().toString();
+            } catch (Exception ex) {
+                return location;
+            }
+        }
     }
 
     private String getParseErrorMessage(String originalMessage, String location) {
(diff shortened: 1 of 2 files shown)


seonwooj0810:
Bumping this — fixes a regression resolving 3.1 refs when the local path contains spaces. Let me know if the normalization approach needs adjusting.

PR #2367, opened 2026-07-10
fix: resolve $ref inside allOf of unreferenced component schemas

### LINKED ISSUE #2270: [Bug]: allOf reference not resolved fully
## Description

Setting `resolveFully(true)` does not resolve `$ref` pointers inside `allOf` schemas. The `$ref` remains as a pointer.

This is while Parsing OpenAPI 3.0.x specs. In case it's relevant, the `allOf` I'm trying to parse contains a `$ref` AND a raw schema as the 2nd element.

## Affected Version
- 2.1.22 (latest)

## Steps to Reproduce

Full reproducible example here: https://github.com/julianmclain/swagger-parser-allOf-resolution
 
Set `ParseOptions.setResolve(true)` and `setResolveFully(true)` and parse a spec with a schema using `allOf`:
```java
String spec = """
    {
      "openapi": "3.0.3",
      "info": { "title": "Test", "version": "1.0" },
      "paths": {},
      "components": {
        "schemas": {
          "SObject": {
            "type": "object",
            "properties": {
              "attributes": { "type": "object" }
            }
          },
          "Contact": {
            "allOf": [
              { "$ref": "#/components/schemas/SObject" },
              {
                "type": "object",
                "properties": {
                  "FirstName": { "type": "string" }
                }
              }
            ]
          }
        }
      }
    }
    """;

options.setResolve(true);
options.setResolveFully(true);

SwaggerParseResult result = new OpenAPIV3Parser().readContents(spec, null, options);
Schema<?> contact = result.getOpenAPI().getComponents().getSchemas().get("Contact");

System.out.println(Json.pretty(contact.getAllOf()));
```

## Expected Behavior

The `$ref` is resolved to the referenced schema content, i.e.

```json
[ {
  "type" : "object",
  "properties" : {
    "attributes" : {
      "type" : "object"
    }
  }
}, {
  "type" : "object",
  "properties" : {
    "FirstName" : {
      "type" : "string"
    }
  }
} ]
```

## Actual Behavior

The reference to "#/components/schemas/SObject"` is unresolved:

```json
[ {
  "$ref" : "#/components/schemas/SObject"
}, {
  "type" : "object",
  "properties" : {
    "FirstName" : {
      "type" : "string"
    }
  }
} ]
```

## Environment
- Java version: 21
- Build tool: Gradle 9.0
- OS: macOS

## Additional Context

There were related issues, but they were marked resolved so I'm not sure if they're relevant: #1157, #1161, #1538

## Checklist
- [x] I have searched the existing issues and this is not a duplicate.

seonwooj0810:
Fixes #2270

### Root cause
`ResolverFully.resolveFully(OpenAPI)` copies the component maps into its internal fields but then only iterates `openAPI.getPaths()` to drive resolution. Any component schema that is **not reachable through a path** is therefore never resolved — so a `$ref` nested inside the `allOf` of such a component schema stays unresolved even when `resolveFully(true)` is set (as reported, and confirmed by a second user).

### Change
After walking the paths, also resolve the captured component schemas. `resolveSchema(...)` is backed by an `IdentityHashMap` cache and an in-progress guard, so component schemas already resolved while walking the paths are returned as-is — no double resolution, and recursive/self-referential schemas are safe.

### Tests
Added `Issue2270Test` + `issue-2270/openapi.yaml`: a spec with empty `paths` and a component `Contact = allOf: [ $ref SObject, {inline} ]`. The test asserts that after `resolveFully(true)` the inner `$ref` is resolved (gone, with `SObject`'s properties inlined). It **fails before** this change (`expected null, but was: #/components/schemas/SObject`) and **passes after**.

### Verification done
- Reproduced the bug on master with the new test (fails before the fix, passes after).
- Ran the full `swagger-parser-v3` module test suite: **610 tests, 0 failures, 0 errors** (`mvn -pl modules/swagger-parser-v3 test`).

### Note for reviewers
Previously-unreferenced component schemas now get resolved by `resolveFully`, which is the intended meaning of the option, but it is a visible output change for any consumer that relied on such components staying in raw `$ref`/`allOf` form. Not covered by any prior test, so flagging it explicitly.

### DIFF
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/ResolverFully.java
@@ -133,6 +133,19 @@ public class ResolverFully {
                 resolvePath(pathItem);
             }
         }
+
+        // Resolve component schemas that are not reachable through any path, so that
+        // e.g. a $ref inside an allOf of an otherwise-unreferenced component schema is
+        // still fully resolved. resolveSchema caches its results, so component schemas
+        // already resolved while walking the paths are returned as-is (no double resolution).
+        if (schemas != null) {
+            for (String schemaName : new ArrayList<>(schemas.keySet())) {
+                Schema resolved = resolveSchema(schemas.get(schemaName));
+                if (resolved != null) {
+                    schemas.put(schemaName, resolved);
+                }
+            }
+        }
     }
 
     public void resolvePath(PathItem pathItem){
--- a/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/Issue2270Test.java
+++ b/modules/swagger-parser-v3/src/test/java/io/swagger/v3/parser/test/Issue2270Test.java
@@ -1,0 +1,42 @@
+package io.swagger.v3.parser.test;
+
+import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.parser.OpenAPIV3Parser;
+import io.swagger.v3.parser.core.models.ParseOptions;
+import io.swagger.v3.parser.core.models.SwaggerParseResult;
+import org.junit.Test;
+
+import java.util.List;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class Issue2270Test {
+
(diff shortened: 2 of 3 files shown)


seonwooj0810:
Following up here — worth noting the description flags a visible behavior change for previously-unreferenced component schemas, so wanted to make sure that's seen before review.

PR #2382, opened 2026-08-18
fix: External Refs with same name are ignored (#2055)

### LINKED ISSUE #2055: External Refs with same name are ignored
When external referenced components have the same name but different definitions, only the first is considered, the rest seem to be ignored. This seems to be because of the way the JSON pointers are used to derive the component name in [Refutils](https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/util/RefUtils.java#L43).

**RefUtils.java**
```
   if (definitionPath != null) { //the name will come from the last element of the definition path
```
"the name will come from the last element of the definition path" - I couldn't find any documentation as to why this is done, nor did I see any mention of this in [this spec](https://spec.openapis.org/oas/v3.0.3#reference-object). 
This could likely cause issues as the definitions with same names are allow if the it is provided in-line.

**P.S:** Originally identified by @crankydillo.

**openapi.json**
```
{
    "openapi": "3.0.3",
    "info": {
        "version": "1.0.0",
        "title": "OpenAPI Service",
        "description": "Sample OpenAPI generated swagger contract"
    },
    "paths": {
        "/a-path": {
            "post": {
                "operationId": "ignoredRefPost",
                "parameters": [
                    { "$ref": "./external_ref_1.json#/limit" }
                ],
                "requestBody": { "$ref": "external_ref_1.json#/a-request" },
                "responses": {
                    "200": {
                        "$ref": "external_ref_1.json#/a-response"
                    },
                    "default": {
                        "$ref": "#/components/responses/default-string"
                    }
                }
            },
            "put": {
                "operationId": "ignoredRefPut",
                "parameters": [
                    { "$ref": "./external_ref_2.json#/limit" }
                ],
                "requestBody": { "$ref": "external_ref_2.json#/a-request" },
                "responses": {
                    "200": {
                        "$ref": "external_ref_2.json#/a-response"
                    },
                    "default": {
                        "$ref": "#/components/responses/default-string"
                    }
                }
            }
        }
    },
    "components": {
        "responses": {
            "default-string": {
                "description": "some string",
                "content": {
                    "text/plain": {
                        "schema": {
                            "type": "string"
                        }
                    }
                }
            }
        }
    }
}
```

**external_ref1.json**
```
{
  "limit": {
    "name": "limit",
    "description": "There can be only 1!",
    "in": "query",
    "schema": {
      "type": "integer",
      "maximum": 1,
      "minimum": 0
    }
  },
 "a-request": {
   "required": true,
   "description": "a string req",
   "content": {
     "text/plain": {
       "schema": {
         "type": "string"
       }
     }
   }
},
  "a-response": {
    "description": "some string response",
    "content": {
      "text/plain": {
        "schema": {
          "type": "string"
        }
      }
    }
  }
}
```

**external_ref_2.json**
```
{
  "limit": {
    "name": "limit",
    "description": "There can be only 2!",
    "in": "query",
    "schema": {
      "type": "integer",
      "maximum": 2,
      "minimum": 0
    }
  },
  "a-response": {
    "description": "some string response",
    "content": {
      "appilication/json": {
        "schema": {
          "schema": {
            "$ref": "#/an-object"
            }
        }
      }
    }
  },
  "an-object": {
    "type": "object",
    "properties": {
      "a-prop": {
        "type": "string"
      }
    }
  },
  "a-request": {
    "required": true,
    "description": "a JSON req",
    "content": {
      "application/json": {
        "schema": {
          "$ref": "#/an-object"
        }
      }
    }
  }
}
```
**Main.java**
```
options.setResolve(true);
OpenAPI openAPI = new OpenAPIV3Parser().read("openapi.json", null, options);
System.out.println("Parameters: " + openAPI.getComponents().getParameters().keySet());
System.out.println("Responses: " + openAPI.getComponents().getResponses().keySet());
System.out.println("Requests: " + openAPI.getComponents().getRequestBodies().keySet());
```

ewaostrowska:
# Pull Request

## Description
When two external files each define a component with the same local name (e.g., both expose a schema called `limit`), the parser was silently discarding the second definition and pointing both usages at the first. This caused incorrect, incomplete resolved documents that were hard to diagnose because no error or warning was emitted. The same problem applied to all component types (schemas, parameters, responses, request bodies, headers, security schemes, links, examples, callbacks).

## What was broken
**Same-name external components were merged into one, losing data:**

```java
// POST /a-path references external_ref_1.json#/limit  ("There can be only 1")
// PUT  /a-path references external_ref_2.json#/limit  ("There can be only 2")

// Before: both operations resolved to the same component; second definition was silently dropped
openAPI.getComponents().getParameters().size() == 1  // wrong — only "limit" present

// After: each distinct external ref gets its own key
openAPI.getComponents().getParameters().size() == 2  // ✓  "limit" + "limit_1"
post.getParameters().get(0).getDescription()  // "There can be only 1" ✓
put .getParameters().get(0).getDescription()  // "There can be only 2" ✓
```
---

## What changed

- **`ExternalRefProcessor`** — replaced the ad-hoc, per-type name-collision logic with a single call to `ComponentNameAllocator.allocate(...)`. The old code had a different (buggy) implementation duplicated for every component type; the new code is a single loop that appends `_N` suffixes until a free or reusable slot is found.

- **`ComponentNameAllocator` (new class)** — centralises name allocation for all OpenAPI component maps. Uses pluggable policies for key lookup (exact vs. case-insensitive), ref-equivalence (delegated to `ResolverCache`), and value-equality reuse. Schemas use case-insensitive key lookup (matching existing behaviour); all other component types use exact-case keys.

---
Fixes:  https://github.com/swagger-api/swagger-parser/issues/2055

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] ♻️ Refactor (non-breaking change)
- [ ] 🧪 Tests
- [ ] 📝 Documentation
- [ ] 🧹 Chore (build or tooling)

## Checklist

- [x] I have added/updated tests as needed
- [x] I have added/updated documentation where applicable
- [x] The PR title is descriptive
- [x] The code builds and passes tests locally
- [x] I have linked related issues (if any)

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,10 @@ final OpenAPI openAPI = new OpenAPIV3Parser().read("a.yaml", null, parseOptions)
 
 This applies to schemas, parameters, responses, pretty much everything containing a ref.
 
+For OpenAPI 3.0 documents, the `OpenAPIResolver` and `ExternalRefProcessor` path adds numeric suffixes such as `Name_1` for name collisions.
+This behavior prevents data loss. It can change generated component keys, local `$ref` values, and component counts for documents with name collisions.
+OpenAPI 3.1 uses a separate dereferencer. This numeric-suffix change does not modify OpenAPI 3.1 resolution.
+
 #### 2. resolveFully:
 
 ```java
--- a/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ComponentNameAllocator.java
+++ b/modules/swagger-parser-v3/src/main/java/io/swagger/v3/parser/processors/ComponentNameAllocator.java
@@ -1,0 +1,70 @@
+package io.swagger.v3.parser.processors;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+
+import io.swagger.v3.parser.ResolverCache;
+import org.slf4j.Logger;
+
+/**
+ * Allocates unique names for resolved OpenAPI components.
+ *
+ * <p>The caller supplies the policies which are specific to a component type:
+ * key lookup, reference equivalence, and reuse of an already resolved value.</p>
+ */
+final class ComponentNameAllocator {
+
+    private final Logger logger;
+    private final ResolverCache cache;
+
+    ComponentNameAllocator(Logger logger, ResolverCache cache) {
+        this.logger = logger;
+        this.cache = cache;
+    }
+
+    <T> String allocate(Map<String, T> components, String baseName,
+            T incoming, String incomingRef, Function<T, String> refOf) {
+        return allocate(components, baseName, incoming, incomingRef, refOf,
+                name -> components.containsKey(name) ? name : null,
+                (existing, ignored) -> false);
+    }
+
+    <T> String allocate(Map<String, T> components, String baseName,
+            T incoming, String incomingRef, Function<T, String> refOf,
+            Function<String, String> keyOf, BiPredicate<T, T> reusePolicy) {
+        for (int suffix = 0; ; suffix++) {
+            String candidate = suffix == 0 ? baseName : baseName + "_" + suffix;
+            String existingKey = keyOf.apply(candidate);
+            if (existingKey == null) {
+                return candidate;
+            }
+            if (canReuse(components.get(existingKey), incoming, incomingRef, refOf, reusePolicy)) {
+                return existingKey;
+            }
+            logger.debug("A different component already claims the name {}", existingKey);
+        }
+    }
+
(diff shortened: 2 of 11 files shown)


ewaostrowska:
The solution maintains the different approach to the key sensitivity in between schema and non-schema components. 

For **schema component**:
`Pet` and `pet` can be resolved to the same component

For `non-schema component`:
`Pet` and `pet` won't be resolved to the same component

I have decided to keep this separation as this would be a breaking change for the users and therefore if such appraoch is confirmed I would prefer to see it in a major version.

djankows (approved):

